PR #6, opened 2016-05-20
Extracting a library from existing cli 

che85:


fedorov:
Once the dcmtk fix is in master, I will update the dcmtk package and merge this. Let me know when dcmtk is ready.

fedorov:
build error:  https://ci.appveyor.com/project/fedorov/dcmqi/build/0.0.110#L1007

apps/seg/ImageSEGConverter.h
```diff
@@ -0,0 +1,90 @@
...
+namespace dcmqi {
+
+    class ImageSEGConverter {
+
+    public:
+        static int itkimage2dcmSegmentation(vector<string> dicomImageFileNames, vector<string> segmentationFileNames,
+                                             const char *metaDataFileName, const char *outputFileName);
+
+        static int dcmSegmentation2itkimage(const char *inputSEGFileName, const char *outputDirName);
+        static int CHECK_COND(const OFCondition& condition);
```
fedorov: @che85 as I work on the SR converter, I realized CHECK_COND should probably remain pound-defined. The reason is when it is a function (even if the function is declared inline), `__LINE__` will not be initialized to the location where it was called. I will move it out into a macro definition, unless you have other suggestions.

che85: No that's fine. Just move it and I will adapt the other code to it.

PR #22, opened 2016-06-24
19 ipfs testing

pieper:
WIP: do not merge.  This is for discussion.

This adds a cliwrap python utility to support running the programs with url inputs.

The utility is used to run a test where the data is pulled from a url.

The url comes from ipfs.  If ipfs proves useful, cliwrap can be extended to directly use ipfs to copy and serve test data, but this part requires more testing and discussion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,4 +250,11 @@ set(TEMP_DIR ${CMAKE_BINARY_DIR}/Testing/Temporary)
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/jsoncpp)
 set(JSONCPP_SRCS ${CMAKE_CURRENT_SOURCE_DIR}/jsoncpp/jsoncpp.cpp)
 
+# utilities
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/util/cliwrap.py
+  ${DCMQI_INSTALL_BIN_DIR}
+  COPYONLY
+  )
+
 add_subdirectory("apps")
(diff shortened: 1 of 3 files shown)


fedorov:
@pieper thanks! I am now getting to this one.

As I was doing the background checks, I found this project http://dat-data.com/, which appears to be rather similar to ipfs [1], but seems to be more active. Did you know about it?

Not sure if this is a right strategy, but as a general rule I prefer to use tools that have support and active community (dat seems to have a lot more favorable status in terms of recent commits, total commits, number of committers).

In any case, I will check out your contribution, but just wanted to let you know about this yet another option!

[1] https://github.com/ipfs/faq/issues/119

pieper:
Yes, Luca pointed me at dat when he visited.  I didn't try it yet.

I think I mentioned in one of the commits or wikis but Hans and I didn't
find ipfs to be especially stable in our testing.  That's why the pull
request is more arms-length in terms of using just a url but not assuming
it is ipfs.

Where I'd like to get to is to have the cliwrap script be able to handle
various data sources and destinations.  Probably DICOMweb is a natural one
to use, but others could be used as well.

fedorov:
ok, I will play with dat tomorrow. I agree it will be nice to have a generic way to handle data sources, but my immediate priority is to organize the test data. I tried this PR, and ran into troubles very quickly

image

I will look into dat tomorrow ...

fedorov:
image

and there is no response ...

pieper:
I restarted the ipfs daemon on the ec2 machine and the url works for me now.

But in general this is the issue with ipfs I mentioned earlier, that it's
not robust enough for general use.

If you look at the cliwrap command line in the test you can try swapping in
a url from anywhere.  Right now it's pretty simple, but we can discuss what
other features we want (like interacting with a the data-data command line
or something).

fedorov:
we decided not to pursue the IPFS approach for test/sample data

PR #23, opened 2016-06-24
Added color picker

che85:


che85:
@fedorov Merging PR since you are on your way to Boston

Issue #17, opened 2016-06-22
Mixing up JSON-LD and JSON Schema seems not to work correctly like it is structured right now


che85:
[...]

PR #24, opened 2016-06-28
JSON-LD/schema refactoring

fedorov:
Not for merging - to be discussed

work to address #17 and #25


fedorov:
@che85 let's discuss this tomorrow

fedorov:
on a separate note, tomorrow we should also look into why tests are failing on travisci, but not on the other 2!

fedorov:
reference for the complex schema and usage of $ref: https://spacetelescope.github.io/understanding-json-schema/structuring.html

fedorov:
@che85 I decided to merge this with the pm schema work branch for #25, as they are all related. I think this should now be ready for review. Let me know what you think.

doc/seg-schema.json
```diff
@@ -3,81 +3,49 @@
...
-      "properties" : {
-        "CodeValue" : { "type" : "SH", "default" : "T-D0050"},
-        "CodingSchemeDesignator" : { "type":"SH", "default" : "SRT"},
-        "CodeMeaning" : { "type":"LO", "default" : "Tissue"}
-      },
-      "required" : ["CodeValue","CodingSchemeDesignator","CodeMeaning"]
-    }
-  },
-
   "seriesAttributes": {
```
che85: seriesAttributes should be SeriesAttributes

che85:
@fedorov Do you also plan to add an example JSON file for parametric maps?

Other than that it looks good to me.

doc/seg-schema.json
```diff
@@ -0,0 +1,135 @@
...
+            {"default" : "SEMIAUTOMATIC"}
+          ],
+          "required" : true,
+        },
+
+        "SegmentAlgorithmName": {
+          "$ref" : "http://qiicr.org/schema/dcmqi/common-schema.json#/definitions/LO"
+        },
+
+        "recommendedDisplayRGBValue": {
```
che85: recommendedDisplayRGBValue should be RecommendedDisplayRGBValue

doc/seg-schema.json
```diff
@@ -0,0 +1,135 @@
...
+          "type": "array",
+          "minItems" : 3,
+          "maxItems" : 3,
+          "items" : { "$ref" : "http://qiicr.org/schema/dcmqi/common-schema.json#/definitions/US"},
+          "default" : ["128","174","128"],
+          "required" : false
+        }
+
```
che85: Missing , after }

doc/seg-schema.json
fedorov: No, this is intentionally lowercase, since recommendedDisplayRGBValue is not a DICOM concept. CIELab is. I am following this convention to differentiate the two types of attributes.

doc/seg-schema.json
fedorov: Fixed!

fedorov:
Great idea re PM example. I will add it now.

fedorov:
@hmeine 

> With "comment", I just meant something like a regular // JS comment (are those valid in JSON?)

No, I don't think this is a valid JSON syntax. What I saw in other documents is to have "description" attribute, but the concern is that this will split the definition between in-schema "description", and the pointer defined by JSON-LD, so I prefer not to do it. Yes, wiki or alike documentation should be good.

PR #49, opened 2016-07-29
ENH: add support for multiple source instances per seg frame

### LINKED ISSUE #47: Support multiple source images per segmentation slice
This is needed when segmentation is done on a dynamic series, for example.

fedorov:
- lists of image instances from the input DICOM file list that match origin
 of the segmentation slice are maintained
 - instances that are not matchin are skipped

fixes #47 (at least, for the BMMR use case)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,8 +4,6 @@ os:
 clone_depth: 1
 skip_tags: true
 clone_folder: c:\dcmqi
-install:
-  - cinst cmake
 configuration:
   - Release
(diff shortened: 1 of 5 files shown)


fedorov:
appveyor is failing for other reason

@che85 let me know if you have any concerns

che85:
That's strange. But it worked before, right?

fedorov:
they did some infrastructure upgrades. With the latest commit, appveyor is back to life!

Issue #20, opened 2016-06-23
Add support for conversion to and from DICOM parametric maps


che85:

fedorov:
Sample dataset and JSON added in #24

fedorov:
@che85 can you tell me what is the status of this? is the "writer" ready, or there are things that remain to be done?

by the way, David Clunie told me that OsiriX supports parametric map, so it should be a good experiment to see how it handles our output (whenever it is ready).

che85:
In my opinion it should be ready, but the code needs to be adapted to most recent changes in dcmtk pmap.

I'll create a PR with mentioning the dcmtk pmap working revision.

PR #53, opened 2016-08-01
Implementation of parametric map itk to dcm converter

che85:
fixes partially #20

works with revision 07e5b0d91b3e0c3dc25419e57343b1ca7c3b8fa1


fedorov:
when you say "partially" - what is missing?

che85:
from DICOM to itk is missing

fedorov:
ok, thanks - just wanted to check if "to DICOM" is complete

fedorov:
Items not populated properly (probably because they are assigned defaults):

* DERIVED\PRIMARY\VOLUME\MTT - probably the last item should be something else (I think MTT = Mean Transit Time), I think we talked about it, but I need to return to that discussion for details and recall what I suggested
* FrameAnatomy/Laterality
* RescaleSlope is set to 1e-6, but 1 in the output
* LUTLabel should probably be something different from "Just testing"
* did you initialize RealWorldValueLastValueMapped?
* quantity is initialized incorrectly
image
* LUTExplanation should not be "We are mapping trash to junk."
* @michaelonken is this right that RWVM module is replicated in the per-frame sequence? It is identical across frames? Is this a shortcut because you don't have the comparison implemented? If so, does it make sense to give an option to the user to decide what should be shared and what per-frame in FGs?

che85:
I took from "DERIVED\PRIMARY\VOLUME\MTT" https://github.com/michaelonken/dcmtk/blob/paramaps/dcmpmap/tests/make_pmp.cc#L85

I am not sure if more information should go into meta information or how much we can automatically populate
https://github.com/che85/dcmqi/blob/paramaps/lib/ParaMapConverter.cpp#L30-L33

Also see for RealWorldValueLastValueMapped and RescaleSlope 
https://github.com/che85/dcmqi/blob/paramaps/lib/ParaMapConverter.cpp#L191-L195

LUTExplanation should maybe go to meta information?

fedorov:
@michaelonken nevermind that comment addressed to you earlier, this is probably outside of dcmtk. Sorry I didn't have time to look at the source code of the converter in more detail yesterday.

fedorov:
Resulting object displayed in OsiriX:

image

che85:
Nice!

che85:
@fedorov looks good to me. Ready to merge if you want.

fedorov:
Next what I want to do (before passing this for checks and review by @dclunie) is to add an option to import patient/study context from the source series.

fedorov:
CI is doing some serious work helping with debugging platform specific issues, nice! If only we could do something similar with Slicer.

che85:
Yes I agree!

fedorov:
:+1:

PR #57, opened 2016-08-05
Add segmentation contexts in JSON based off "AIM on ClearCanvas" lists

### LINKED ISSUE #42: Prepare terminology files in JSON
It probably makes sense to use JSON for all tasks.

fedorov:
Resolves #42


fedorov:
@che85 the idea is to use `doc/segContexts/SegmentationCategoryTypeModifierRGB.json` and `doc/segContexts/AnatomicRegionAndModifier.json` to populate the corresponding sections of the web app.

Later we will have more specialized contexts, and the user would choose which one to use.

PR #61, opened 2016-08-07
Reorganize library

fedorov:
followup on #58, rebased to the current master; aims to resolve #50

PR #63, opened 2016-08-08
Reorganize library

msmolens:
This builds on #61 by adding a commit to fix the executable link errors on Linux builds.


fedorov:
@msmolens thanks for the fix - but the windows dashboard is still failing, which started after I added executable only flag. Do you know what is the problem?

fedorov:
Dashboards are green - thanks a lot 👍 @msmolens! Can you tell me how were you able to update the commit without creating a new branch and new PR? The initial commit was failing the tests, but you fixed it with subsequent commits.

msmolens:
@fedorov Great! Here, I updated the commit by amending it and then force pushing the branch to github. This keeps the commit history cleaner once merged. For example, on my branch:

   &lt;edit CMakeLists.txt&gt;
    git add CMakeLists.txt
    git commit --amend
    git push --force

An alternative is to use `git rebase` to squash commits.

But, in general you don't want to force push to a public branch, as it rewrites the history. It is safe in this case where the branch is in my own fork and no one else can write to it.

fedorov:
@msmolens good to know! thanks for the hint, and for the fixes!

PR #65, opened 2016-08-17
Add handling of multiple labels per input image file

### LINKED ISSUE #44: Communication of files + label IDs should be fixed
When ITK labels are converted to DICOM, there may be multiple labels per single ITK label file. Therefore, we need a combination of file (or itkImageData) and label ID to do the conversion properly.

Currently, this is not reflected in the SEG schema, and probably not in the API.

Open question is how file list should be communicated in the JSON. Options:

1. Pass file list outside of JSON, have groups of segment groups, where each group maps to the position of the input label file, and within each group labelID maps to the individual label value within that file. This maps more cleanly into the API use case, since segment groups can the same way map to the position of the itkImageData in the vector passed to the API.

2. Include file name directly in the segment group. To be used with the API, without file IO, JSON will need to be modified before it is passed downstream.

fedorov:
Supposed to resolve #44

### DIFF
--- a/doc/seg-schema.json
+++ b/doc/seg-schema.json
@@ -1,195 +1,198 @@
 {
-  "@context": "http://qiicr.org/dcmqi/contexts/dcmqi.jsonld",
-  "id": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/seg-schema.json#",
-  "$schema": "http://json-schema.org/draft-04/schema#",
-  "properties": {
-    "seriesAttributes": {
-      "type": "object",
-      "properties": {
-        "ContentCreatorName": {
-          "allOf": [
+  "@context":"http://qiicr.org/dcmqi/contexts/dcmqi.jsonld",
+  "id":"https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/seg-schema.json#",
+  "$schema":"http://json-schema.org/draft-04/schema#",
+  "properties":{
+    "seriesAttributes":{
+      "type":"object",
+      "properties":{
+        "ContentCreatorName":{
+          "allOf":[
             {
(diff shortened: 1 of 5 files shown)


fedorov:
@che85 let me know if you have concerns

note that this PR will require changes to the webapp to handle lists of segment attributes

che85:
@fedorov I worked on the schema as well and made corrections based on the validation tools. It would make more sense to first merge my changes and adapt your changes to it...

fedorov:
sure, are they very extensive? go ahead with the PR

I am working on another PR to add jsonlint testing

fedorov:
rebased to current master

doc/seg-schema.json
```diff
@@ -1,198 +1,201 @@
...
+  "$schema":"http://json-schema.org/draft-04/schema#",
+  "properties":{
+    "seriesAttributes":{
+      "type":"object",
+      "properties":{
+        "ContentCreatorName":{
+          "allOf":[
             {
-              "$ref": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/common-schema.json#/definitions/PN"
+              "$ref":"https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/common-schema.json#/definitions/PN"
```
che85: @fedorov Why are you removing all those spaces? It's general json syntax to have 
```
"key": "value" 
```

Even when you use a library for generating JSON it always comes out that way. Maybe it's a matter of taste but let's try keep it one way. 

http://www.jsonschemavalidator.net/ also formats the way I mentioned.

fedorov: It's just what the formatter I used generated: https://jsonformatter.curiousconcept.com/

I will run in http://www.jsonschemavalidator.net/ and update

che85: Oh didn't know that you are also using one :)

PR #66, opened 2016-08-17
Fixed json schemata

che85:
validation based on http://www.jsonschemavalidator.net/

### DIFF
--- a/doc/common-schema.json
+++ b/doc/common-schema.json
@@ -1,6 +1,6 @@
 {
   "@context": "http://qiicr.org/dcmqi/contexts/dcmqi.jsonld",
-  "id" : "http://qiicr.org/schema/dcmqi/common-schema.json",
+  "id" : "https://raw.githubusercontent.com/che85/dcmqi/fixSchema/doc/common-schema.json",
   "$schema": "http://json-schema.org/draft-04/schema#",
 
   "definitions" : {
--- a/doc/common-schema.json
+++ b/doc/common-schema.json
@@ -46,9 +46,9 @@
       "type" : "object",
       "required" : ["CodeValue","CodingSchemeDesignator","CodeMeaning"],
       "properties" : {
-        "CodeValue" : { "type" : "SH", "default" : "T-D0050"},
-        "CodingSchemeDesignator" : { "type":"SH", "default" : "SRT"},
-        "CodeMeaning" : { "type":"LO", "default" : "Tissue"}
+        "CodeValue" : { "$ref": "#/definitions/SH", "default" : "T-D0050"},
+        "CodingSchemeDesignator" : { "$ref": "#/definitions/SH", "default" : "SRT"},
+        "CodeMeaning" : { "$ref": "#/definitions/LO", "default" : "Tissue"}
       }
     }
   }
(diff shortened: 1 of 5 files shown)


che85:
@fedorov  will need to change the reference links

doc/common-schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "@context": "http://qiicr.org/dcmqi/contexts/dcmqi.jsonld",
-  "id" : "http://qiicr.org/schema/dcmqi/common-schema.json",
+  "id" : "https://raw.githubusercontent.com/che85/dcmqi/fixSchema/doc/common-schema.json#",
```
fedorov: point to a location under qiicr/dcmqi instead?

che85: That's what I just commented on my last comment for the PR ;)

doc/seg-schema.json
```diff
@@ -1,152 +1,197 @@
...
-        "SegmentDescription": { "type" : "string", "required" : false},
-
-        "SegmentedPropertyCategoryCodeSequence": {
-          "allOf" : [
-            {"$ref" : "http://qiicr.org/schema/dcmqi/common-schema.json#/definitions/codeSequence"},
-            {"properties" :
-             {"CodeValue" : {"default": "T-D0050"},
-              "CodingSchemeDesignator" : {"default": "SRT"},
-              "CodeMeaning" : {"default": "Tissue"}}
+  "properties": {
```
fedorov: why is this "properties" here?

doc/seg-schema.json
```diff
@@ -1,152 +1,197 @@
...
+          ],
+          "default": "Reader1"
+        },
+        "ClinicalTrialSeriesID": {
+          "allOf": [
+            {
+              "$ref": "https://raw.githubusercontent.com/che85/dcmqi/fixSchema/doc/common-schema.json#/definitions/LO"
             }
           ],
-          "required" : true
```
fedorov: why did you remove required? is it redundant if there is a default?

doc/seg-schema.json
```diff
@@ -1,152 +1,197 @@
...
+              "26565",
+              "37722"
+            ]
+          }
         }
       }
     }
   }
-}
+}
```
fedorov: github doesn't like that you don't have newline in the end ;-)

doc/seg-schema.json
che85: Without it, the schema is not valid

doc/seg-schema.json
```diff
@@ -1,152 +1,197 @@
...
+    "segmentAttributes": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": [
+          "SegmentedPropertyCategoryCodeSequence",
+          "SegmentedPropertyTypeCodeSequence",
+          "SegmentAlgorithmType"
+        ],
+        "properties": {
```
che85: @fedorov all required attributes moved into one array on top of properties, not for each separate

fedorov: oh, ok

LabelID should also be required

che85: done

doc/seg-schema.json
fedorov: oh, I see, I missed it - I thought it is present only in this schema, but it also shows up for pm-schema. Alright

fedorov:
Looks good! I am merging, since I don't expect anything new on CI for json changes

PR #72, opened 2016-08-22
BUG: fixed compiling with referenced schemata

che85:
- added tests for example seg, pm and sr.
- fixed typo

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -25,19 +25,19 @@ set(AJV_OPTIONS --all-errors)
 
 foreach(json_common_schema_file ${json_common_schema_files})
   # -r needs to be used when specifying a schema that has references to external files
-  list(APPEND REF_SCHEMATA -r ${json_common_schema_file})
+  list(APPEND common_schema_references -r ${json_common_schema_file})
 endforeach()
 
 foreach(json_common_schema_file ${json_common_schema_files})
   get_filename_component(json_name ${json_common_schema_file} NAME_WE)
   add_test(NAME ajv_${json_name}
-    COMMAND ajv compile -s ${json_common_schema_file} ${REF_SCHEMATA} ${AJV_OPTIONS})
+    COMMAND ajv compile -s ${json_common_schema_file} ${common_schema_references} ${AJV_OPTIONS})
 endforeach()
 
 foreach(json_schema_file ${json_schema_files})
   get_filename_component(json_name ${json_schema_file} NAME_WE)
   add_test(NAME ajv_${json_name}
-           COMMAND ajv compile -s ${json_schema_file} ${REF_SCHEMATA} ${AJV_OPTIONS})
+           COMMAND ajv compile -s ${json_schema_file} ${common_schema_references} ${AJV_OPTIONS})
 endforeach()
 
 foreach(json_example_file ${json_example_files})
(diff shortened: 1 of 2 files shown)


doc/CMakeLists.txt
```diff
@@ -2,18 +2,50 @@ file(GLOB json_files
...
-  add_test(NAME ajv_${json_name}
-    COMMAND ajv compile -s ${json_schema_file})
-endforeach()
+set(BASELINE ${CMAKE_SOURCE_DIR}/doc/)
+set(COMMON_SCHEMA ${BASELINE}/common-schema.json)
+set(SR_COMMON_SCHEMA ${BASELINE}/sr-common-schema.json)
+
+set(AJV_OPTIONS --all-errors)
+
+get_filename_component(json_name common-schema.json NAME_WE)
```
fedorov: This is redundant. I used it to extract file name from the full path, which is no-op here.

doc/CMakeLists.txt
```diff
@@ -2,18 +2,50 @@ file(GLOB json_files
...
+get_filename_component(json_name pm-example.json NAME_WE)
+add_test(NAME ajv_${json_name} COMMAND ajv -s ${BASELINE}/pm-schema.json
+  -r ${COMMON_SCHEMA}
+  -d ${BASELINE}/${json_name}.json)
+
+get_filename_component(json_name sr-tid1500-example.json NAME_WE)
+add_test(NAME ajv_${json_name} COMMAND ajv -s ${BASELINE}/sr-tid1500-schema.json
+  -r ${COMMON_SCHEMA}
+  -r ${SR_COMMON_SCHEMA}
+  -d ${BASELINE}/${json_name}.json ${AJV_OPTIONS})
```
fedorov: why "all erors" here, but not for other tests?

fedorov:
Do you think this can be rewritten to have a loop for generating the tests (to simplify)?

What happens if you pass a schema which is not reference with `-r` to ajv? If nothing bad happens, maybe we can just pass all schemas to all tests to simplify?

Can you add some comments (lines starting with `#`) to explain what are the things you are testing, and what is the difference between the tests (some have `-d` flags). I am lazy to read the docs :)

che85:
Oh yeah I wanted to add --all-errors to all tests, forgot to add them for the others. Will add them.

I don't know how to add a loop for that case. 

I will try with adding all common schemata

che85:
@fedorov What do you think about the most recent commit?

doc/CMakeLists.txt
```diff
@@ -2,50 +2,48 @@ file(GLOB json_files
...
+  get_filename_component(json_name ${json_schema_file} NAME_WE)
+  add_test(NAME ajv_${json_name}
+           COMMAND ajv compile -s ${json_schema_file} ${REF_SCHEMATA} ${AJV_OPTIONS})
+endforeach()
+
+foreach(json_example_file ${json_example_files})
+  get_filename_component(json_name ${json_example_file} NAME_WE)
+  STRING(REGEX REPLACE -example.json -schema.json json_schema ${json_example_file})
+  add_test(NAME ajv_${json_name}
+    # -d stands for data that needs to be validated
```
fedorov: :) what I meant is to add a comment for each group of tests, explaining what is going on, not just a comment on one flag!

looks great otherwise!

fedorov:
👍

che85:
Ah ok will fix this. What about the bmmr-example? It's going to fail anyway, because there is no schema defined...

fedorov:
bmmr should follow seg schema. Do you want to add missing ref,or should I do it after this PR is merged?

che85:
Let's do it after this PR is merged...

PR #73, opened 2016-08-22
ENH: reorganize the tests into groups by object type

fedorov:
The motivation is that we will have multiple tests for each object type. When a
new test is added, its name would just need to be added to the corresponding group.


che85:
@fedorov do you think that we still need jsonlint? ajv also throws syntax errors

fedorov:
Good point. I think the difference is that jsonlint is run for all json files, and it makes sense to run it for all, while ajv tests are run specifically for json schemas and instantiation. So I think it still makes sense to keep jsonlint (for example, we also use json for terminology, and I know there will be more uses later, for similar purposes).

fedorov:
Doesn't add much time to testing anyway.

PR #76, opened 2016-08-22
ENH: json data/schema validation, download button, default values from schema

che85:
- using requirejs + angularjs now 
- download button provided for downloading json output as file
- delay for triggering validation only when user stops typing in output textarea
- using default values from seg-schema.json

TODO: we really need to think about situations, where the user has no internet:
- in that caste no seg context will be loaded
- no validation schema will be loaded

try at: http://che85.github.io/dcmqi/#/seg


fedorov:
"Body substance" item looks different - is this expected?

image

fedorov:
This still does not support selection of "segmentation context" selector. Is it because you are not sure how to do that, or are you waiting for me to provide more than one context?

fedorov:
> we really need to think about situations, where the user has no internet

let's for now have this as a limitation, and revisit after RSNA

che85:
I am waiting for you to provide more.

fedorov:
What is the process for updating the web app if a schema is updated?

che85:
image

fedorov:
makes sense!

che85:
Usually the schema is taken from master for validation, so it should work fine unless you change the whole structure or names, since those are partially mapped to the form components.

fedorov:
I understand validation, but what if there is a change that would affect the UI controls?

fedorov:
so it would need to be updated manually, I guess. Not criticizing, just asking

che85:
Yes, since we don't expect the UI to ask for all attributes. Even when we would ask for all, we still would need to set the context for SegmentedCategoryType/Modifier etc. 

It's und impossible but for now I would leave it that way

PR #80, opened 2016-08-25
Add SPL Generic Atlas context

fedorov:


fedorov:
@lassoan @cpinter: this file https://github.com/fedorov/dcmqi/blob/ece877cee871d67b83c74dbb51c5e885095e439c/doc/segContexts/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json adds segmentation context that parallels SlicerGenericAnatomy (a subset of the codes in the DICOM "master list"). I also added with this commit additional attribute to keep the name of the label used in the Slicer LUT, as we discussed.

PR #82, opened 2016-08-25
context selectors for SegmentationCategoryType and AnatomicRegion

### LINKED ISSUE #43: Add an option to the SEG web app to support dictionary selection
After #42 is resolved

che85:
fixes #43

- also added drop area for auto populating information with DICOM input file

see http://che85.github.io/dcmqi


fedorov:
It is confusing that category/type selector is grayed out (but not disabled!) when the page is loaded - is this intentional?

image

fedorov:
I suggest moving Anatomic region context selector inside the "Optional" section where anatomic region/modifier selectors from the context are located. What do you think?

image

fedorov:
If there is a new context JSON file - what is the procedure to make the web app aware of it? Would I just drop it alongside the existing ones under gh-pages branch?

fedorov:
I would also comment out the part of the code that adds the DICOM file drop area - since this functionality is mostly not used, it only eats space, and can add to the confusion. Users can also be concerned that we want to take their DICOMs and send them off to Google! ;-)

che85:
That is on purpose. Angular material handles it that way to show, that the attributes has not been defined yet.

Assuming that it would not be greyed out, the user could think that it is already populated.
> It is confusing that category/type selector is grayed out (but not disabled!) when the page is loaded - is this intentional?
> 
> image
>

che85:
I am not sure about that since it is defined "generally" for the webapp and not per segment. 

In any case I would leave it out of the Segment Attributes section

> I suggest moving Anatomic region context selector inside the "Optional" section where anatomic region/modifier selectors from the context are located. What do you think?
> 
> image

che85:
For now you would need to extend the following list:
https://github.com/che85/dcmqi/blob/093b5bda9e29bea17d1ee2dc0ee4a8b0c01c5b2a/js/app.js#L14-L17

I will try to find a solution for automatically detecting it. 

> If there is a new context JSON file - what is the procedure to make the web app aware of it? Would I just drop it alongside the existing ones under gh-pages branch?

fedorov:
> That is on purpose. Angular material handles it that way to show, that the attributes has not been defined yet.

Interesting. But why is NOT Anatomic region grayed out for consistnecy?

Also, when selection of anatomic region is not allowed, the selector is also not greyed out (but in fact is disabled, as in the example below).

image

fedorov:
> For now you would need to extend the following list:
https://github.com/che85/dcmqi/blob/093b5bda9e29bea17d1ee2dc0ee4a8b0c01c5b2a/js/app.js#L14-L17

This is simple enough, I don't think auto detection is important.

che85:
Because I added a check that if only one context is selected, it should automatically select that one.

che85:
Oh wait, let me check something. I see what you mean now.

che85:
That's interesting. On my laptop it looks like this:
image

fedorov:
I am now not sure which comment you refer to! I think this is how it looks on mine too!

che85:
1. I am disabling the SegmentedCategoryContext selector for further selection:
  - as soon as the user selects a SegmentedCategory and 
  - if there are more than 1 segment is defined.

It does not make sense to change the context once the user started ....

fedorov:
> I am disabling the SegmentedCategoryContext selector for further selection:
> as soon as the user selects a SegmentedCategory or
> if there are more than 1 segment is defined.
> I am disabling the AnatomicRegionContext selector for further selection:
> as soon as the user selects a AnatomicRegion or
> if there are more than 1 segment is defined
> It does not make sense to change the context once the user started ....

yes, this I understand, and I agree. 

I think all of this makes sense, although it is a bit confusing. But it doesn't need to be perfect on the first merge.

Thanks!

PR #83, opened 2016-08-25
Cleanup tid1500writer

fedorov:


apps/sr/tid1500writer.cxx
```diff
@@ -105,15 +98,21 @@ int main(int argc, char** argv){
...
+          ff.getDataset()->findAndInsertCopyOfElement(commonTagsToCopy[t],&imageLibGroupDataset);
+        }
+        CHECK_COND(report.getImageLibrary().addImageEntryDescriptors(imageLibGroupDataset));
       }
 
-      OFCHECK(report.getImageLibrary().addImageGroup().good());
-      OFCHECK(report.getImageLibrary().addImageEntry(imageLibDataset).good());
-      OFCHECK(report.getImageLibrary().addImageEntryDescriptors(imageLibDataset).good());
+      DcmDataset imageEntryDataset;
+      DcmTag imageTagsToCopy[] = {DCM_Modality,DCM_SOPClassUID,DCM_SOPInstanceUID,DCM_ImagePositionPatient};
```
fedorov: @jriesmeier this is the remaining issue that does not work as expected - if I don't include `DCM_Modality` in the line above, no descriptors show up for image entries:

image

 If I do include it, modality is duplicated in the image entry descriptors (it is already included at the group level):

image

I hope it is again me not doing something properly, and not a bug!

jriesmeier: A simple fix would be not to copy the Modality element to the imageLibGroupDataset. But, I guess you want to have that on group-level (with the other common values) and not on the image-level.
Omitting it from the imageEntryDataset will also not work as the value of the Modality element is used to determine which modality-specific sub-template (TID 1603 to 1607) to use. That means, the API of the TID 1600 would need to be extended in this regard in order to support this use case.

jriesmeier: I simple solution for extending the TID class would be to add another enum to TID1600_ImageLibrary::AddImageMode. Actually, this is what this type was intended for. Currently, it only supports two values: withoutDescriptors and withAllDescriptors. I could easily add withAllDescriptorsButModality :) Alternatively, I could introduce another variant of addImageEntry() that expects the Modality value as a parameter (and not as part of the dataset).

fedorov: It sounds like in order to address the immediate needs, I will need to operate on the SR tree and probably just manually remove the modality entries from the per-image descriptors, right?

jriesmeier: Yes, this will also work. Navigating through the tree should be easy since all content items that belong to a template row are annotated by "TID xxxx - Row n". You can use gotoAnnotatedNode() and gotoNextAnnotatedNode() for this purpose. See test case "dcmsr_TID1001_ObservationContext". Of course, XPath expressions would also be nice... :)

fedorov: I have this code that works
```C++
  // cleanup duplicate modality from image descriptor entry
  DSRDocumentSubTree st = report.getImageLibrary().getTree();
  size_t nnid = st.gotoAnnotatedNode("TID 1601 - Row 1");
  st.goDown();
  st.removeSubTree();
  while(nnid){
    nnid = st.gotoNextAnnotatedNode("TID 1601 - Row 1");
    st.goDown();
    st.removeSubTree();
  }
  st.print(cout);
```
but `getTree` returns a const reference, so I cannot modify the content.

Is there a way to go around this?

My next resort is to operate on `DSRDocumentTree`, instead of the `DSRDocumentSubTree`, but then navigation becomes more complicated.

fedorov:
> I would recommend to use DcmTagKey instead of DcmTag for the array (since it does not require a dictionary lookup for the proper VR). Furthermore, you should declare this "variable" const. The same applies to "imageTagsToCopy".

done

apps/sr/tid1500writer.cxx
jriesmeier: The const reference is on purpose: the user should not manipulate the internally managed template subtree directly. However, after you've called "doc.setTreeFromRootTemplate()" you should be able to access the DSRDocumentTree returned by "doc.getTree()"... DSRDocumentTree is derived from DSRDocumentSubTree, so gotoNextAnnotatedNode() et al. are also available.

fedorov: Ok, will try this next ...

While I was waiting, I realized multiple inheritance allows me to do the following, but no method to remove element!
```C++
  TID1600_ImageLibrary &il = report.getImageLibrary();
  //DSRDocumentSubTree st = il.getTree();
  size_t nnid = il.gotoAnnotatedNode("TID 1601 - Row 1");
  il.gotoChild();
  //il.removeSubTree();
  while(nnid){
    nnid = il.gotoNextAnnotatedNode("TID 1601 - Row 1");
    il.gotoChild();
    //il.removeSubTree(); <- no I can't!
  }
  il.print(cout);
  cout << " --- " << endl;
  TID1600_ImageLibrary &il1 = report.getImageLibrary();
  il1.print(cout);
```

fedorov: > The const reference is on purpose: the user should not manipulate the internally managed template subtree directly. However, after you've called "doc.setTreeFromRootTemplate()" you should be able to access the DSRDocumentTree returned by "doc.getTree()"... DSRDocumentTree is derived from DSRDocumentSubTree, so gotoAnnotatedNode() et al. are also available.

Whooh, yes - this worked!

> DSRDocumentTree is derived from DSRDocumentSubTree

I overlooked this, and for whatever reason assumed it is the other way, so didn't think about it!

Thank you!

jriesmeier: > While I was waiting, I realized multiple inheritance allows me to do the following, but no method to remove element!

Correct, removing content items is disabled in the SR (sub)template classes since it would/could break the internally managed structures.

fedorov:
@jriesmeier I think I addressed all of your concerns, so I am going to merge unless you object. Thanks for all the help with refining this!

apps/sr/tid1500writer.cxx
jriesmeier: Two further comments: Instead of gotoChild() I would suggest to go to the "Modality" child content item using gotoNamedChildNode(), which is more robust IMHO. Also the code duplication (calling gotoChild() and removeSubTree() twice, before and within the while loop) could be avoided but is probably OK for a temporary workaround like this.

fedorov:
I resolved the remaining issue in the last commit

apps/sr/tid1500writer.cxx
```diff
@@ -190,22 +190,23 @@ int main(int argc, char** argv){
...
-  st.removeSubTree();
-  while(nnid){
-    nnid = st.gotoNextAnnotatedNode("TID 1601 - Row 1");
-    st.goDown();
-    st.removeSubTree();
+  {
+    DSRDocumentTree &st = doc.getTree();
+    DSRCodedEntryValue modalityCode("121139","DCM","Modality");
+    size_t nnid = st.gotoAnnotatedNode("TID 1601 - Row 1");
+    nnid = st.gotoNamedChildNode(modalityCode);
```
jriesmeier: You can directly use CODE_DCM_Modality, which is defined in "codes/dcm.h"

apps/sr/tid1500writer.cxx
```diff
@@ -190,22 +190,23 @@ int main(int argc, char** argv){
...
-  st.goDown();
-  st.removeSubTree();
-  while(nnid){
-    nnid = st.gotoNextAnnotatedNode("TID 1601 - Row 1");
-    st.goDown();
-    st.removeSubTree();
+  {
+    DSRDocumentTree &st = doc.getTree();
+    DSRCodedEntryValue modalityCode("121139","DCM","Modality");
+    size_t nnid = st.gotoAnnotatedNode("TID 1601 - Row 1");
```
jriesmeier: You don't check the returned node ID of gotoAnnotatedNode() but as I said: this is just a workaround and if it works for your use case... it's should be sufficient :)

apps/sr/tid1500writer.cxx
```diff
@@ -190,22 +190,23 @@ int main(int argc, char** argv){
...
+    nnid = st.gotoNamedChildNode(modalityCode);
+    while(nnid){
+      CHECK_COND(st.removeSubTree());
+      nnid = st.gotoNextAnnotatedNode("TID 1601 - Row 1");
```
jriesmeier: Same as above.

apps/sr/tid1500writer.cxx
```diff
@@ -197,14 +195,19 @@ int main(int argc, char** argv){
...
-    nnid = st.gotoNamedChildNode(CODE_DCM_Modality);
-    while(nnid){
-      CHECK_COND(st.removeSubTree());
-      nnid = st.gotoNextAnnotatedNode("TID 1601 - Row 1");
-      nnid = st.gotoNamedChildNode(CODE_DCM_Modality);
+  //  - if we have any imageLibrary items supplied
+  if(metaRoot.isMember("imageLibrary")){
+    if(metaRoot["imageLibrary"].size()){
+      DSRDocumentTree &st = doc.getTree();
+      size_t nnid = st.gotoAnnotatedNode("TID 1601 - Row 1");
```
jriesmeier: What about:
`      size_t nnid = st.gotoAnnotatedNode("TID 1601 - Row 1");`
`      while (nnid) {`
`        nnid = st.gotoNamedChildNode(CODE_DCM_Modality);`
`        if (nnid) {`
`          CHECK_COND(st.removeSubTree());`
`          nnid = st.gotoNextAnnotatedNode("TID 1601 - Row 1");`
`        }`
`      }`

fedorov:
yes, this is nicer, I committed

Next time, feel free to fork and contribute directly, so your contribution to the code can be recorded!

Issue #54, opened 2016-08-02
Establish a mechanism to communicate key parameters


fedorov:
Specific example: how to communicate which b-values were used for model fit?

Awaiting suggestions from @dclunie

dclunie:
If one is encoding an ADC map as an MR or Enhanced MR or Parametric Map IOD, then one can use the RWVM Quantity Definition Sequence to encode one or more items, one of which is the Quantity itself, and others of which can describe the method of calculation and/or its (numeric) parameters, and would could encode the numeric parameters to describe the b-values of the original images. E.g., for an ADC map calculated from B0 and B1000 source images using a mono-exponential model one might encode something like:

(G-C1C6, SRT, "Quantity") = (113041, DCM, "Apparent Diffusion Coefficient")
(G-C036, SRT, "Measurement Method") = (xxxx0, 99XXX, "Monoexponential ADC from log of ratio of two samples")
(xxxx1, 99XXX, "Source image diffusion b-value") = 0
(xxxx1, 99XXX, "Source image diffusion b-value") = 1000

I have pre-coordinated the concepts of both the "source image" and the "diffusion b-value" into one code to use as the concept name of the numeric parameter, because we don't have a way to post-coordinate these to allow re-use of a code for just "b-value"; arguably we could just omit the "source image" component of the concept and just send "b-value" and the recipient could assume it was an input parameter (in the general case, this would reduce the combinatorial explosion of pre-coordinated codes).

Arguably one could omit mention of the b-value = 0 parameter, but in reviewing the old literature, I see that in the classic Burdette paper [1], a b-value of 1, not 0, was used.

Even some of the more complicated methods of computing ADC maps, such as these that involve different B values on a voxelwise basis, could be represented this way  by listing all the collected b-values, even though one could not indicate which were used for which voxel (which would require another map). E.g., we might need methods such as:

"Monoexponential ADC from least squares fit of multiple samples"
"Monoexponential ADC from voxelwise selection of b-value"

See also [3].

Anyway, we should probably write a DICOM CP to add standard codes for the various methods and parameters.

We should probably consult with an MR physicist like Tom Chenevert and/or the QIBA MRI PDF group, to prepare a good list of methods and make sure we have captured the necessary parameters.

Before I forget to mention in, we also need to send the correct units in Measurement Units Code Sequence, which in UCUM are "mm2/s" (assuming all scaling issues have been appropriately dealt with [4], as should be the case for images you are creating).

David

[...]

PR #85, opened 2016-09-11
Improvements to PM converter

fedorov:
- added the capability to communicate relevant parameters, see #54 
- composite context is now populated to have a complete output object
- improved how slope is initialized to simplify interpretation of the result


fedorov:
Merging this, with the understanding that related issues #88 and #90 remain unresolved ...

PR #100, opened 2016-10-19
BUG: fixed issues with json validation (issue #98):

### LINKED ISSUE #98: Schema conformance validation appears to be not reliable
`recommendedDisplayRGBValue` is lower case in the schema

https://github.com/QIICR/dcmqi/blob/master/doc/seg-schema.json#L162

and upper case in this instantiation

https://github.com/QIICR/dcmqi/blob/master/doc/bmmr-example.json#L34

but the dashboard is green!

che85:
fixes #98 
- prohibiting  additional properties
- needed to define @schema which is usually not used in json data files
- removed MeasurementGroup keyword since this name is used within definitions. Measurements already defines an array of objects which hold those values.
- corrected some typos regarding CodeSequences
- using number instead of string for measurement values how it was defined in the schema

### DIFF
--- a/apps/sr/tid1500writer.cxx
+++ b/apps/sr/tid1500writer.cxx
@@ -50,14 +50,6 @@ DSRCodedEntryValue json2cev(Json::Value& j){
     j["CodeMeaning"].asCString());
 }
 
-template <typename T>
-string numberToString(T number)
-{
-  ostringstream sstream;
-  sstream << number;
-  return sstream.str();
-}
-
 void addFileToEvidence(DSRDocument &doc, string dirStr, string fileStr){
   DcmFileFormat ff;
   OFString fullPath;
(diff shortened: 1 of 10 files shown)


fedorov:
The issue with using numbers instead of strings is that it leads to a potential inconsistency due to number representation. If you read 1.23 and then save it back, it may not be the same number, since we are not prescribing precision. I am not certain switching from string to number is a good idea.

che85:
I can change it back. then value would be a string. I am fine with that. I only changed it because you defined value as a number in json and that's why i got errors while validating.

fedorov:
@pieper do you have an opinion on how measurements should be communicated into the converter - as strings, or as numbers?

pieper:
If possible we should use the same definitions of numbers in string form
that DICOM uses (e.g. DS)

http://dicom.nema.org/medical/dicom/current/output/html/part05.html#sect_6.2

Maybe we can include a regular expression for this:

"0"-"9", "+", "-", "E", "e", "." and the SPACE character of Default
Character Repertoire

che85:
sounds good. I will look into this and fix it

che85:
@fedorov Ready for merge unless you don't have any comments. I will rebase your PR after that.

fedorov:
:+1:!

pieper:
+1

fedorov:
> Ready for merge unless you don't have any comments.

http://www.ef.com/english-resources/english-grammar/how-use-unless/

che85:
Yes I know.... I shouldn't use unless unless I know how to use it ;)

fedorov:
That's how you learn!

Issue #97, opened 2016-10-12
Compare two json files for round-trip testing


fedorov:
To improve testing, it makes sense to "round trip" tests for the metadata, in a similar way as we have round-trip tests for image data.

To help with this task, we could add a tool that takes two jsons, and tests whether they are identical.

fedorov:
motivated by #96

fedorov:
https://github.com/monsur/jsoncompare/blob/master/jsoncompare.py

fedorov:
this one looks more capable, since we will need to skip certain keys, more popular, and more recently updated: https://github.com/ChannelIQ/jsoncompare

PR #101, opened 2016-10-20
Roundtrip test and some fixes to inconsistencies between the schema, examples, and JSONMetadataHandler

che85:
Followup on #97.

jsoncompare is from https://github.com/ChannelIQ/jsoncompare, with ChannelIQ/jsoncompare#8 applied.

### DIFF
--- a/apps/seg/Testing/CMakeLists.txt
+++ b/apps/seg/Testing/CMakeLists.txt
@@ -5,6 +5,9 @@ add_test(NAME ${MODULE_NAME}_hello
 
 set(BASELINE ${CMAKE_SOURCE_DIR}/data/ct-3slice/)
 
+set(MODULE_TEMP_DIR ${TEMP_DIR}/seg)
+make_directory(${MODULE_TEMP_DIR})
+
 include_directories(${ITK_INCLUDE_DIRS})
 add_executable(${MODULE_NAME}Test ${MODULE_NAME}Test.cxx )
 target_link_libraries(${MODULE_NAME}Test ${${MODULE_NAME}_TARGET_LIBRARIES})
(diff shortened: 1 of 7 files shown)


che85:
@fedorov Please check if all your commits (except of the last one for adding the missing file) are there.

fedorov:
Did you expect the test would fail too? It sounded from the earlier communication that you expect they should pass, so I wanted to confirm.

che85:
I only wanted to correct the syntax of the test. I didn't mean or check anything else. But I can go ahead and check what's wrong.

fedorov:
ah, ok - makes sense. Can you take a look how much effort would be to fix inconsistencies? I didn't plan to merge until the test passes.

che85:
@fedorov I figured out, that the DICOM-attributes of https://github.com/QIICR/dcmqi/blob/master/data/ct-3slice/liver.dcm
don't even correspond to the attributes defined in https://github.com/QIICR/dcmqi/blob/master/doc/seg-example.json

Instead of using the liver.dcm from the repository, we should use the output from itkimage2segimage otherwise the roundtrip would not make sense because we don't continue working with the produced data, right?

fedorov:
Good observation! It probably makes sense to run itk2seg first using baseline metadata, then run seg2itk on the output of itk2seg, not baseline seg, and use the output metadata file from seg2itk to compare with the baseline json that was used in the itk2seg in the first step. Does this make sense?

che85:
I propose to do the following:
- run itkimage2segimage with liver_seg.nrrd, 01.dcm..., seg-example.json
  - produces liver.dcm
- run segimage2itkimage with product liver.dcm
  - produces nrrd file and meta.json
- comparing meta files by means of comparejson.py with input1 seg-example.json and input2 meta.json

I would set dependencies between the tests in order to have the right order when running them.

che85:
makes sense ;)

che85:
@fedorov ready for merge

fedorov:
Looks great! One question - can you write up a summary of the changes you did to jsoncompare, and consider submitting pull request to the upstream?

che85:
@fedorov I already submitted a PR to the upstream just before I created the PR here. 

https://github.com/ChannelIQ/jsoncompare/pull/9

fedorov:
Very good - I didn't check that repo, so thanks for adding a reference here.

che85:
In my last commit message I added a reference to your PR and my PR of jsoncompare upstream

fedorov:
Sorry, I missed it

PR #103, opened 2016-10-24
Removed key "seriesAttributes" and moved all underlying keys to root level

### LINKED ISSUE #102: Remove seriesAttributes key from schema, example and converter, web interface
@fedorov I think we discussed about that in the past. We wanted to remove the seriesAttributes key and have all sub-keys on the same level like segmentAttributes. 

It would look more like:
https://github.com/QIICR/dcmqi/blob/master/doc/pm-example.json
or
https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-example.json

Webapp priority set to low.

che85:


fedorov:
let's see if tests clear, but looks ok to me

che85:
Shall I update the webapp as well or is that not priority right now?

fedorov:
if it is non-trivial - not a high priority

che85:
Ok will wait for approving this PR and then might take a look at the webapp because its validation schema is the master...

che85:
@fedorov: passed

fedorov:
👍 looked good to me from the quick look

PR #112, opened 2016-11-09
Add support for build apps option

jcfr:
Per our discussion, this topic adds option `BUILD_APPS`

It also:
* ensure that the `BUILD_PARAMETRIC_MAPS` is passed to the inner build
* add `SlicerExecutionModel` as a dependency only if `BUILD_APPS` is `ON`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,10 @@ cmake_dependent_option(
 mark_as_advanced(BUILD_PARAMETRIC_MAPS)
 mark_as_superbuild(BUILD_PARAMETRIC_MAPS)
 
+option(BUILD_DOC "Build ${PROJECT_NAME} documentation." ON)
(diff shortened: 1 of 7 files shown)


jcfr:
Cc: @fedorov @nolden

fedorov:
@jcfr if implementing this feature is more difficult than expected, I would postpone working on it.

The most important issue is to have dcmqi integrated with Slicer. I think @nolden has a workaround, and we can look at this issue after dcmqi integration is done.

Also see relevant issue reported today by @che85, which will be a stopper for the Reporting work: http://slicer-devel.65872.n3.nabble.com/Extension-Manager-doesn-t-ask-for-installing-dependencies-td4037570.html.

So it looks like there are more critical issues to look into. What do you think?

jcfr:
I am merging, the last change affect only appveyor and the CI was previously green for Travis and CircleCI

PR #123, opened 2016-11-11
Mprage test

pwighton:
Hi dcmqi team,

Here are a couple of modest enhancements that have helped me with development/debugging. One is a flag to allow segimage2itkimage to output niftis, the other is a flag force itkimage2segimage to not skip empty slices.  It looks like there already was a similar flag that has been removed, so I'll understand if you're not keen on putting it back in.

-Paul

### DIFF
--- a/apps/seg/itkimage2segimage.cxx
+++ b/apps/seg/itkimage2segimage.cxx
@@ -33,7 +33,7 @@ int main(int argc, char *argv[])
   ifstream metainfoStream(metaDataFileName.c_str(), ios_base::binary);
   std::string metadata( (std::istreambuf_iterator<char>(metainfoStream) ),
                        (std::istreambuf_iterator<char>()));
-  DcmDataset* result = dcmqi::ImageSEGConverter::itkimage2dcmSegmentation(dcmDatasets, segmentations, metadata);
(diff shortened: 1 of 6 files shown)


che85:
@pwighton Just a heads up: your local email address for git is not connected to your github account here. There is no proper user displayed for your commits.

pwighton:
Thanks @che85!  I'll try to fix that this weekend.

fedorov:
@pwighton thanks for the PR!

About the first one - I wonder if it makes sense to have the functionality more generic, and allow specifying any of the meaningful output types supported by ITK - I can see how some people might want to have .mhd, for example. This could be done by specifying a set of supported file types, and have NRRD as default (similar to how parameter enumeration is defined here: https://github.com/Slicer/Slicer/blob/master/Modules/CLI/ResampleScalarVolume/ResampleScalarVolume.xml#L22-L37). What do you think?

About skipping empty slices - it definitely makes sense to bring it back. What I am not sure if how well tested that functionality is. Did you check that mode of operation, does it work for you?

pwighton:
Sounds good @fedorov.  I can work on expanding the output types later this week.  Initial tests with --noskip are looking good!  I have a roundtrip test where a niftii file goes in with a PIL orientation and comes out with a PIR orientation but I think the fov they both express are equivalent.

fedorov:
@pwighton up to you if you want to close this PR and work on a new one, or refine this PR!

pwighton:
I'll close this one, fix my github account issues, and make a new one.

PR #127, opened 2016-11-14
BUG: cmake/SuperBuild: Ensure svn and git paths are propagated

jcfr:
This commit fixes issue #126 by completing the patch initially integrated in
b3ea5fa (BUG: cmake/SuperBuild: Only propagate svn and git paths if valid values)

PR #128, opened 2016-11-14
ENH: ci: Trigger appveyor/travis builds only with PR or master branch updates

jcfr:


jcfr:
@fedorov  This should address https://github.com/QIICR/dcmqi/pull/127#issuecomment-260494914

fedorov:
it is interesting that we only had these 3 builds until recently, I think I had them set up in the CI settings directly, but something got reset...

Thank you!

PR #130, opened 2016-11-15
BUG: enable parametric maps component

fedorov:
Fix consistency of the cmake flag, update paramap CLI macro parameters, make
parametric maps enabled by default.

It will make sense to remove parametric map flag completely. The reason we had
it because that component of dcmqi was being developed while parametric map API
was not yet integrated into DCMTK. Now that it is part of the stable DCMTK API,
there is no real need to make parametric map optional.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,7 @@ option(DCMQI_BUILD_APPS "Build ${PROJECT_NAME} applications." ON)
 mark_as_superbuild(DCMQI_BUILD_APPS)
 
 cmake_dependent_option(
-  DCMQI_BUILD_PARAMETRIC_MAPS "Build parametric maps for ${PROJECT_NAME}." OFF
+  DCMQI_BUILD_PARAMETRIC_MAPS "Build parametric maps for ${PROJECT_NAME}." ON
   "DCMQI_BUILD_APPS" OFF
   )
 mark_as_superbuild(DCMQI_BUILD_PARAMETRIC_MAPS)
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.5.0)
 
-if(BUILD_PARAMETRIC_MAPS)
+if(DCMQI_BUILD_PARAMETRIC_MAPS)
   add_subdirectory(paramaps)
 endif()
 add_subdirectory(seg)
--- a/apps/paramaps/CMakeLists.txt
+++ b/apps/paramaps/CMakeLists.txt
@@ -32,23 +32,35 @@ set(COMMON_SOURCE_FILES
 #-----------------------------------------------------------------------------
 set(MODULE_NAME itkimage2paramap)
 
+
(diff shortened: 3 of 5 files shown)


fedorov:
@che85 do you know if we ever built parametric maps component on Windows? It looks like it was always disabled by default (this is the flag before changes of JC): https://github.com/QIICR/dcmqi/blob/9022e5a57a47a47bb26a80b5525dd0fcc3938d70/CMakeLists.txt#L72. Many many errors on appveyor! Just wondering if you tried to build on Windows.

che85:
We disabled it because by the time parametric maps was not implemented in dcmtk as far as I remember. We never built it for Windows no.

fedorov:
I cannot make sense of this error on appveyor:

```
[00:01:02]      4>C:\DCMTK-install\include\dcmtk/dcmiod/cielabutil.h(198): error C2062: type 'double' unexpected [C:\dcmqi\dcmqi-build\dcmqi-build\apps\paramaps\itkimage2paramapLib.vcxproj] [C:\dcmqi\dcmqi-build\DCMQI.vcxproj]
```

this is the failing line: http://git.dcmtk.org/?p=dcmtk.git;a=blob;f=dcmiod/include/dcmtk/dcmiod/cielabutil.h;h=781477b51664cba25260563fb73ef59c5d16a255;hb=cece99c797380dc859f7aeb878cd2deeef7a5e3d#l198

che85:
I guess there is maybe just some include missing. We had that situation in the past. Let me check.

fedorov:
yes, I also recall you had this issue in the past, but I don't remember much more than that!

che85:
Me neither. I thought that there was a PR, where you described that issue, but I don't know which one. There are so many ....

fedorov:
@michaelonken do you have any ideas about the [Windows errors](https://ci.appveyor.com/project/fedorov/dcmqi/build/0.0.378) below? Maybe this looks familiar to you.

image

fedorov:
@michaelonken - as Christian said - super-confusing, but this fixed the Windows build error ... Looks like a dcmtk issue to me, maybe you might want to look into this. For now, I am spared from setting up Windows build environment from scratch to debug this! whooh....

jcfr:
Within visual studio, if you try to "follow" the definition of "double" where does it bring you ?

Look like double is already defined somewhere ...

fedorov:
@jcfr yes, it seemed to me as well that this might be a `#define` issue. But I could not find any for either double or min in either dcmqi or dcmtk. "following" the definition is not feasible for me - the only visual studio I have is the one running on appveyor!

msmolens:
It could be a conflict between `static double min(double a, double b);` in that file and the Windows definition of the macro called `min`.

fedorov:
Ah, so min IS a `#define`d macro on Windows!

fedorov:
Merging this, since Travis build passed earlier, and since it takes so long to get build scheduled, it is better to save it for master branch ...

michaelonken:
Ok, I will rename the method in DCMTK then!

PR #132, opened 2016-11-16
BUG: add option to enable testing

fedorov:


che85:
@fedorov This line says[1] enable_testing. Actually I would expect it to work then. And at least I would expect the option to appear in cmake.

[1] https://github.com/QIICR/dcmqi/blob/master/CMakeLists.txt#L267

fedorov:
@che85 I wouldn't know how cmake treats those, but as you can see from appveyor, looks like we needed to explicitly define `ENABLE_TESTING`, right?

fedorov:
superseded by #133

PR #136, opened 2016-11-16
Display message if ajv or jsonlint is missing

jcfr:


jcfr:
@fedorov  The last commit could be revisited, it current display a status message if ajv and/or jsonlint can not be found. Since `BUILD_DOC` allow to skip adding this directory, we could also error out asking the user to reconfigure with `-DBUILD_DOC:BOOL=OFF`

doc/CMakeLists.txt
```diff
@@ -15,18 +19,22 @@ file(GLOB json_schema_files
   )
 list(REMOVE_ITEM json_schema_files ${json_common_schema_files})
 
-if(WIN32)
-  set(NPM_SUFFIX ".cmd")
-endif()
```
fedorov: nice!

fedorov:
Thank you JC, this is very nice!

We should also think about reorganization - "doc" is not quite the best name for that directory, and is misleading (and it is my fault!)!

jcfr:
> "doc" is not quite the best name for that directory

Agreed. I wasn't sure why it was named like this. 

What about split into two:
* dcmqi/schema
* dcmqi/schema/examples

I am not sure about `segContexts`, it is not used anywhere.

To explain how to use of the schema, may be something similar to this: https://spacetelescope.github.io/understanding-json-schema/

fedorov:
>> "doc" is not quite the best name for that directory
> Agreed. I wasn't sure why it was named like this.

The original motivation for naming it that way was that those schemas essentially serve as machine-readable documentation of the parameters communicated to the converters. You suggested organization makes sense.

> I am not sure about segContexts, it is not used anywhere.

They are used, but it is not easy to see! :)

First, they are used by the web app (gh-pages branch) to populate terminology selectors: http://qiicr.org/dcmqi/#/seg

Second, they are used by the Segmentation/Terminology infrastructure in Slicer (currently, copies: https://github.com/Slicer/Slicer/blob/4bac13a687b39a69e2648f281bd737bae3eaf08c/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json and https://github.com/Slicer/Slicer/blob/4bac13a687b39a69e2648f281bd737bae3eaf08c/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json).

The idea is that use case specific contexts can be added to this directory, to simplify user interaction (use case specific contexts can consist of 10s of terms, as opposed to 100s or 1000s in the current contexts). Then the idea is that through the community contribution new contexts could be added, and Slicer (if/when dcmqi becomes an external module) and web app could more or less dynamically discover new contexts.

Issue #140, opened 2016-11-16
SEG converter does not work properly with multiple input segmentations


fedorov:
It looks like the problem is in this line: https://github.com/QIICR/dcmqi/blob/master/libsrc/ImageSEGConverter.cpp#L193. What it should probably do is iterate over all labels that are defined in the segmentation attribute list for the position of the given input segmentation label, and find the item that has labelID matching the pixel value for the current segment in the input segmentation file. If no such label is found, it should fail with error.

che85:
fixed

PR #141, opened 2016-11-16
BUG: Creating new reader for each input segmentation file (issue #140)

che85:
- otherwise the same segmentation file will be used over and over again

### DIFF
--- a/apps/seg/itkimage2segimage.cxx
+++ b/apps/seg/itkimage2segimage.cxx
@@ -14,10 +14,10 @@ int main(int argc, char *argv[])
     return EXIT_FAILURE;
   }
 
-  ReaderType::Pointer reader = ReaderType::New();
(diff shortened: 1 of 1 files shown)


fedorov:
this definitely makes sense! Does the comment I have in #140 make sense too, or that is not an issue?

che85:
It works the way it is implemented now. No errors and loading into Slicer afterwards shows the right segmentations

fedorov:
Alright! Sorry for the confusion!

che85:
No confusion here. I also only noticed because I debugged it from the beginning. Otherwise I wouldn't have noticed either.

jcfr:
Would it make sense to update the test to prevent regression like this one in the future ?

fedorov:
it definitely would - we discussed this with Christian earlier today

fedorov:
see #142

Issue #146, opened 2016-11-17
Schema cleanup items


fedorov:
Let's keep adding small things to this issue that should be fixed after RSNA ...

- [ ] `LabelID` -> `labelID`

fedorov:
Also related to this, I started working on the schema for segmentation context related items.

fedorov:
The remaining issues are bigger and need to be reconsidered as part of the revision of the schema organization and harmonization with the standard. This will be done after the 1.0.0 release.

PR #149, opened 2016-11-18
BUG: make capitalization consistent with the converters parameters

fedorov:
see https://github.com/QIICR/Reporting/issues/85#issuecomment-261553027 and #146


fedorov:
merging this since these files are not used directly in dcmqi and should not have any effect on testing

Issue #157, opened 2016-11-23
SEG converter must work when no derivation image items are present


fedorov:
Currently, it segfaults!

fedorov:
Next remaining issue is 

```
E: FrameOfReferenceUID (0020,0052) absent in FrameOfReferenceModule (type 1)
```

@michaelonken where in DCMTK is FrameOfReference initialized for SEG?

In this particular situation, it is missing in the source images, so I will need to do something else (create a new UID probably, although that is not very helpful!), but I also wanted to know what function carries over metadate from the source images into the SEG.

fedorov:
Next issue

```
W: ReferencedInstanceSequence (0008,114a) absent in SeriesAndInstanceReferenceMacro (type 1)
E: ReferencedInstanceSequence (0008,114a) absent in SeriesAndInstanceReferenceMacro (type 1)
Condition failed: W: ReferencedInstanceSequence (0008,114a) absent in SeriesAndInstanceReferenceMacro (type 1)
E: ReferencedInstanceSequence (0008,114a) absent in SeriesAndInstanceReferenceMacro (type 1)
Missing Attribute(s) in /Users/fedorov/github/dcmqi/libsrc/ImageSEGConverter.cpp:385
```

michaelonken:
For the Frame of Reference: You have to set one yourself by getting() the module from the IOD class and calling the specific setter() for the attribute.

For the Referenced Instance Sequence: Can you point me to the code where you populate the Common Instance Reference Module?

PR #158, opened 2016-11-23
BUG: improving the code to handle situations when derivation images a…

fedorov:
…re missing

WIP to address #157 -- DO NOT MERGE!


fedorov:
I am merging this, since I think it is a good improvement over the current functionality, but unfortunately does not resolve the issue completely.

PR #164, opened 2017-01-18
Add seg context schema

fedorov:
related to #146


fedorov:
@che85 can you help me with this one?

I have two issues:

1. I can't figure out why `ajv_anatomic-context-schema` and `ajv_DICOM-Master-anatomiccontext` fail with

```
error: can't resolve reference #ModifierDefinition from id https://raw.githubusercontent.com/fedorov/dcmqi/add-seg-context-schema/doc/segment-context-common-schema.json#
```

`#ModifierDefinition` is present in that file: https://github.com/QIICR/dcmqi/pull/164/files#diff-b13b24ac28c7db3f149677417fdfe0a4R8

2. I don't understand why `ajv_DICOM-Master-segcontext` test succeeds! After the fact, I realized that for the segment context files `Modifier` should be specified as array of items (example: https://github.com/QIICR/dcmqi/blob/master/doc/segContexts/SegmentationCategoryTypeModifier-DICOM-Master.json#L657). The schema expects single item, not array, but validation fails with the array in input! Do you know why? 

Maybe you will see what's going on with a fresh look?

fedorov:
I started working on another issue, and noticed that I didn't have the right path to the definition, should be `#/definitions/ModifierDefinition`. After fixing that, the error is gone, and I get error for the item 2 above. Still don't understand why I was not getting error before. I think there are some rules how validation works in the situation of chained dependencies, and I had wrong expectations ...

che85:
Sorry my mail program didn't notify me about mails received in my sub directories after setting up my new Mac. 

Very nice! LGTM. I am merging this.

PR #170, opened 2017-01-20
Renamed attributes by means of DICOM Dictionary 

che85:
related to #146 

TODO: webapp needs to be adapted to those changes

### DIFF
--- a/include/dcmqi/SegmentAttributes.h
+++ b/include/dcmqi/SegmentAttributes.h
@@ -29,16 +29,17 @@ namespace dcmqi {
     void setRecommendedDisplayRGBValue(const unsigned& r, const unsigned& g, const unsigned& b);
     void setRecommendedDisplayRGBValue(const unsigned rgb[3]);
 
-    void setSegmentedPropertyCategoryCode(const string& code, const string& designator, const string& meaning);
-    void setSegmentedPropertyCategoryCode(const CodeSequenceMacro& codeSequence);
-    void setSegmentedPropertyType(const string& code, const string& designator, const string& meaning);
-    void setSegmentedPropertyType(const CodeSequenceMacro& codeSequence);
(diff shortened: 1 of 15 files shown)


doc/pm-example-float.json
```diff
@@ -22,7 +22,7 @@
     "CodeMeaning": "Mono-exponential diffusion model"
   },
   "SourceImageDiffusionBValues": ["0","1400"],
-  "AnatomicRegionCode": {
+  "AnatomicRegionSequence": {
```
fedorov: I think this one should be `AnatomicRegionCodeSequence`, if we want to be consistent.

include/dcmqi/SegmentAttributes.h
```diff
@@ -60,11 +61,11 @@ namespace dcmqi {
     string segmentAlgorithmType;
     string segmentAlgorithmName;
     unsigned recommendedDisplayRGBValue[3];
-    CodeSequenceMacro* anatomicRegion;
-    CodeSequenceMacro* anatomicRegionModifier;
-    CodeSequenceMacro* segmentedPropertyCategoryCode;
-    CodeSequenceMacro* segmentedPropertyType;
-    CodeSequenceMacro* segmentedPropertyTypeModifier;
+    CodeSequenceMacro* anatomicRegionSequence;
+    CodeSequenceMacro* anatomicRegionModifierSequence;
```
fedorov: `*CodeSequence`

doc/pm-example-float.json
fedorov: Oh, nevermind - you are right! I checked, and it is correct! Interesting ...

include/dcmqi/SegmentAttributes.h
fedorov: never mind!

che85:
Yeah I am also not sure, why it's that inconsistent in DICOM itself ....

Issue #166, opened 2017-01-18
Reorganize doc folder


fedorov:
Need to separate schemas, example metadata JSONs, segmentation context JSONS.

che85:
@fedorov Let me know when you are done with that, so that I can apply changes to the web interfaces.

fedorov:
What do you think about the following structure:

```
doc
 \-- schemas
 \-- segmentationContexts
 \-- examples

**Comment by che85** [comment]
Or something like the following. Either way is fine for me.

```
doc
 \-- pm
   \-- schema.json
   \-- example.json
\-- seg
   \-- schema.json
   \-- example.json
...
```

fedorov:
I prefer the first one, since there will be common components of the schema shared across applications.

che85:
Oh yeah that's right. Agreed!

PR #171, opened 2017-01-20
ENH: reorganize the content of doc folder

fedorov:
Re #166 

I would prefer to merge and fix any things I miss, rather than using temporary URLs in this PR and then fixing them in the master ...


che85:
At least the paths used within the tests like [1] should be fixed 

https://github.com/QIICR/dcmqi/blob/master/apps/paramaps/Testing/CMakeLists.txt#L27

fedorov:
Of course! I missed those. Will do.

Issue #172, opened 2017-01-20
Create docker image


fedorov:
Organization created: https://hub.docker.com/u/qiicr/

fedorov:
Docker image: https://hub.docker.com/r/qiicr/dcmqi/builds/

TODO:
- [ ] update documentation to provide instructions how to install and use from docker container

fedorov:
documentation issue moved to #163

PR #174, opened 2017-01-22
Add integration with docker

fedorov:
This PR adds Dockerfile to the repository (based on https://hub.docker.com/r/thewtex/centos-build/, but using the latest CMake). Since docker image can only have one entry point, the individual command line tools of dcmqi are exposed using a "selector" bash script. Note that `cmake install` does not seem to function as expected (unresolved symbols error while running from the installed location), thus executables are being run from the build tree.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,91 @@
+FROM centos:5
+MAINTAINER http://github.com/fedorov
+
+RUN yum update -y && \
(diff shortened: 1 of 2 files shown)


fedorov:
Resolves #172

fedorov:
Merging, since edits are unrelated to CI.

PR #176, opened 2017-01-23
Support of most popular itk file formats for conversion output from dcm to itk

### LINKED ISSUE #175: Converters should support conversion from DICOM into any itk file format (not only nrrd)
At the moment our converters for DICOM pmap/seg are only expecting a output directory, where image data will always be saved in nrrd file format instead of providing an option of available file formats for storing the image data.

che85:
supported itk output formats: mhd|mha|nii|nifti|hdr|img

fixes #175
related to https://github.com/QIICR/dcmqi/pull/123

@fedorov let me know if I should add other file formats

### DIFF
--- a/libsrc/ImageSEGConverter.cpp
+++ b/libsrc/ImageSEGConverter.cpp
@@ -7,7 +7,8 @@ namespace dcmqi {
 
   DcmDataset* ImageSEGConverter::itkimage2dcmSegmentation(vector<DcmDataset*> dcmDatasets,
                                                           vector<ImageType::Pointer> segmentations,
-                                                          const string &metaData) {
+                                                          const string &metaData,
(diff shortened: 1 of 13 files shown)


libsrc/ImageSEGConverter.cpp
```diff
@@ -167,7 +168,8 @@ namespace dcmqi {
 
         LabelStatisticsType::BoundingBoxType bbox = labelStats->GetBoundingBox(label);
         unsigned firstSlice, lastSlice;
-        bool skipEmptySlices = true; // TODO: what to do with that line?
```
fedorov: what is the reason you commented it out? I think it would be better to test and check this functionality works, and use the flag passed to the function.

PR #178, opened 2017-01-25
BUG: adding nrrd as element instead of having it as default only

che85:
might be related to https://github.com/QIICR/QuantitativeReporting/issues/135

### DIFF
--- a/apps/paramaps/paramap2itkimage.xml
+++ b/apps/paramaps/paramap2itkimage.xml
@@ -33,6 +33,7 @@
       <longflag>--outputType</longflag>
       <description>Output type of the resulting image data.</description>
       <default>nrrd</default>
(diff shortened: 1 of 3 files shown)


fedorov:
Add explicit output type flag for paramap tests as well?

che85:
For all of them or for one only to make sure we cover everything?

fedorov:
I just meant for these 2 tests: https://github.com/QIICR/dcmqi/blob/master/apps/paramaps/Testing/CMakeLists.txt#L44-L81. Only for the sake of consistency, since you added it in the seg test.

che85:
Yeah but I only added it for one seg test so that there is one with and one without the flag

fedorov:
ok, that is fine

Issue #182, opened 2017-01-27
Allow specifying input directory with DICOM instead of list of files


fedorov:
It is very inconvenient to specify the list ...

TODO's:

- [x] add checks for duplicates while loading DICOM instances
- [x] revise the implementation to use DCMTK API for parsing the input directory

fedorov:
Anyone has a solution how to get the list of files in directory in C++?!

fedorov:
https://github.com/cxong/tinydir ?

fedorov:
tinydir solution didn't work for me, used the recipe from http://stackoverflow.com/questions/612097/how-can-i-get-the-list-of-files-in-a-directory-using-c-or-c

fedorov:
TODO:

- [ ] add checks for duplicates while loading DICOM instances

michaelonken:
DCMTK has searchDirectoryRecursively(...), [this is the API](http://support.dcmtk.org/docs/classOFStandard.html#23539eec65d0c9a5e896579a2baf9561).

jriesmeier:
There is "now" also a [version of this function](http://support.dcmtk.org/docs-snapshot/classOFStandard.html#a1fae93f115e75ffec0a7dd6829c74a69) that supports wide chars (based on the OFFilename class).

fedorov:
Thank you! I should definitely use this one.

fedorov:
The option has been added.

TODO:

- [ ] revise the implementation to use DCMTK API for parsing the input directory

che85:
@fedorov I am working right now on this issue

fedorov:
Resolved by #196

PR #183, opened 2017-01-27
ENH: allow passing of reference DICOMs as directory

fedorov:
Also improved error reporting, made DcmDataset loading more robust

re #182

### DIFF
--- a/apps/seg/itkimage2segimage.cxx
+++ b/apps/seg/itkimage2segimage.cxx
@@ -5,12 +5,33 @@
 #undef HAVE_SSTREAM // Avoid redefinition warning
 #include "dcmqi/ImageSEGConverter.h"
 
+#ifdef _WIN32
+#include "dirent_win.h"
+#else
+#include <dirent.h>
+#endif
+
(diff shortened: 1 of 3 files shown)


che85:
Works fine for me. Just tested locally. 

The only thing that I noticed is that the user can specify both:
  --inputDICOMList 
and 
  --inputDICOMDirectory

che85:
For instance if the user specifies:

```
    --inputDICOMList ${BASELINE}/01.dcm,${BASELINE}/02.dcm,${BASELINE}/03.dcm
    --inputDICOMDirectory ${BASELINE}
```

each dataset will be load twice

fedorov:
Thanks!

> user can specify both

I don't know if it's a good idea or not, but it was intentional

che85:
Yeah it's kind of good if files are spread somehow throughout the file system but on the other hand there should be a check, that the same file won't be loaded twice

fedorov:
I agree, I added to the todo list in #182. I need the docker image, that's why wanted to merge quickly...

PR #191, opened 2017-02-01
ENH: using dcmtk searchDirectoryRecursively 

### LINKED ISSUE #182: Allow specifying input directory with DICOM instead of list of files

che85:
fixes #182


che85:
Not sure why the test with recursively searching a inputDICOMDirectory fails on Windows

che85:
Taking a look at the log messages gives me the impression, that a) I am doing something wrong when using OFStandard::searchDirectoryRecursively or b) it doesn't work properly on Windows. @michaelonken, @jriesmeier do you have any idea, what's going on?

fedorov:
@che85 for me the appveyor build page load does not finish. Can you add a screenshot of the error message to this thread?

che85:
@fedorov same here. I downloaded the log file and displayed it in atom. There is no error message. OFStandard::searchDirectoryRecursively just doesn't find any files in the directory, which by another test is invalidated because we are specifying a inputDICOMList with files from that directory.

image

another test where loading succeeds:

image

jriesmeier:
Maybe, it's too easy and too obvious, but Windows uses a "\" as the path separator and not a "/". Are the two screenshots really from a Windows system?

jriesmeier:
Hmm, l've meant a backslash ("\\") in my previous message. Writing from my mobile...

che85:
@jriesmeier the screenshots are from build log. We were having problems opening the appveyor, that's why I downloaded the log and made screenshots....

jriesmeier:
So, the question was: are you using searchDirectoryRecursively() with a slash ("/") as a path separator on a Windows system? This will not work, of course. You have to use a backslash (\\") instead.

che85:
Good point. That was really too obvious. I will commit later. Thanks @jriesmeier

apps/seg/itkimage2segimage.cxx
```diff
@@ -43,12 +45,15 @@ int main(int argc, char *argv[])
   if(dicomDirectory.size()){
     OFList<OFString> fileList;
     cout << "Searching recursively " << dicomDirectory << " for DICOM files" << endl;
+#if _WIN32
+    replace(dicomDirectory.begin(), dicomDirectory.end(), '/', '\\');
```
fedorov: need to confirm there is no OS-independent definition of file separator in dcmtk

jriesmeier:
@fedorov Confirmed. But there is a system-dependent one: PATH_SEPARATOR ;-)

apps/seg/itkimage2segimage.cxx
fedorov: Correction, I misunderstood this code. OS-independent definition of file separator won't help here, but I still wonder if this code can be improved with some DCMTK functionality to avoid `#_WIN32`. @jriesmeier how does DCMTK handle this situation?

fedorov:
@che85 can you factor this functionality out so we can reuse it in in the PM converter?

che85:
Sure!

jriesmeier:
@fedorov We don't have this situation in the DCMTK because if you start e.g. dcmdump on Linux, you have to use a "/", and on Windows a "\\" as the path separator. In our case, it makes no sense to use anything else than the system-specific path separator for specifying directories.

che85:
@fedorov once travis successfully compiled and tested, it's ready for merge

fedorov:
Sorry to bug you, but maybe you could squash some redundant commits?

PR #203, opened 2017-02-20
Standalone packaging rebased

jcfr:


jcfr:
re appveyor failure:

```
At line:1 char:31
+ ... thon27-x64" $env:PATH="$PYTHON_DIR;$PYTHON_DIR/Scripts;$env:PATH" $PA ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Unexpected token '$env:PATH="$PYTHON_DIR;$PYTHON_DIR/Scripts;$env:PATH"' in expression or statement.
At line:1 char:85
+ ... ;$env:PATH" $PACKAGE="C:/dcmqi/dcmqi-build/dcmqi-build/dcmqi-win64.zi ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Unexpected token '$PACKAGE="C:/dcmqi/dcmqi-build/dcmqi-build/dcmqi-win64.zip"' in expression or statement.
At line:1 char:145
+ ... PACKAGE="C:/dcmqi/dcmqi-build/dcmqi-build/dcmqi-win64.zip" If ([strin ...
+                                                                ~~
Unexpected token 'If' in expression or statement.
```

We will have to move the code into its own power shell script ...

jcfr:
@fedorov I updated the topic removing the upload as GitHub release. Now looking into the naming of package for windows.

jcfr:
Et voila. This last commit addresses the windows package names.

I expect CI to turn green

@fedorov Consider also reviewing the change associated with docker, they drastically reduce the size of the images and include build metadata into it. Particularly useful to track provenance of the image.

fedorov:
Great, thank you! I did look at the docker-related changes, but I have to say I don't understand all that is going on or the motivation. I will need to keep looking, and maybe talk with you at some point. I still don't quite get it why we need dockcross, sorry!

jcfr:
> I will need to keep looking, and maybe talk with you at some point. I still don't quite get it why we need dockcross, sorry!

The idea is the following:

* dcmqi docker image is geared toward user, this means should not include a compiler. By basing it only on `centos:5` and copying the binaries into it, we have a light weight image.

*  the binaries copied into it are built using the dockcross image (these are images made to compile project) ... it removes the need for dcmqi project to have to maintain its own image and compilers.

The process is then the following:
1. build the dcmqi binaries using dockcross compiler
2. create a docker image based on centos:5 that only contain the binaries generated in step 1

fedorov:
@jcfr thanks for your patience and explanations. I understand better now. When you have a moment, can you explain what is the current process of updating the docker image? In the past I was able to have automatic build on Docker Hub, and now I see there is a section in CircleCI that is supposed to push the image, but it is not clear to me where it is supposed to get the credentials to complete that step.

fedorov:
> not clear to me where it is supposed to get the credentials to complete that step

I see, ok - found those in CircleCI environment setting. Looking at what happened and why Docker Hub is not updated.

fedorov:
Ok, I think I was confused by the way qiicr/dcmqi docker hub entry is set. It shows build details, but not history for the published images. It does look like the docker push worked. Would be great if we could talk sometime, I feel like I am missing quite a few things!

jcfr:
> When you have a moment, can you explain what is the current process of updating the docker image ? 

Prior to his change, dockerhub was systematically building the docker image independently of test failing or passing. 

In the current implementation, image will be pushed only if tests pass. Not that on circleci, the deployement step is done only for master branch. See [here](https://github.com/QIICR/dcmqi/blob/7f40682681608bfcf62199d563e4b70bca15b945/circle.yml#L29-L37)

>  there is a section in CircleCI that is supposed to push the image, but it is not clear to me where it is supposed to get the credentials to complete that step.

Indeed, the env. variable are set  in the CircleCI UI.  That said, they will be available only for PR done by trusted members. 

secret-not-for-fork-pullrequest

>  I think I was confused by the way qiicr/dcmqi docker hub entry is set. It shows build details, but not history for the published images. It does look like the docker push worked. Would be great if we could talk sometime, I feel like I am missing quite a few things!

The docker automated build are not enabled. See below

docker-hun-auto-build-disabled


But was you can see below, the latest image is updated on each successful build done from master:

dockerhun-image-list

Also worth noting that after pushing a tag, a tagged docker images will also be automatically pushed. This is done [here](https://github.com/QIICR/dcmqi/blob/7f40682681608bfcf62199d563e4b70bca15b945/docker/Makefile#L67)

fedorov:
This helps a lot, thank you! I didn't look under Tags, I was expecting an update image under Build details, but of course it is not supposed to be there. Indeed amazing how much smaller the new image is!

PR #207, opened 2017-02-21
ENH: add version reporting to the converters output

fedorov:
With the current setup, especially with the use of Docker images, it is not easy
to determine the version of the software being used. This commit adds console
output of the git hash and tag (when available) as the first line in the command
line output of the converters.

### DIFF
--- a/CMake/FindGit.cmake
+++ b/CMake/FindGit.cmake
@@ -99,6 +99,15 @@ if(GIT_EXECUTABLE)
           OUTPUT_STRIP_TRAILING_WHITESPACE)
     endif(NOT ${GIT_error} EQUAL 0)
 
+    execute_process(COMMAND ${GIT_EXECUTABLE} describe --tag
+       WORKING_DIRECTORY ${dir}
+       OUTPUT_VARIABLE ${prefix}_WC_TAG
+       OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+    if(NOT ${GIT_error} EQUAL 0)
+      set(${prefix}_WC_TAG "not available")
+    endif(NOT ${GIT_error} EQUAL 0)
+
     execute_process(COMMAND ${GIT_EXECUTABLE} config --get remote.origin.url
        WORKING_DIRECTORY ${dir}
        OUTPUT_VARIABLE ${prefix}_WC_URL
--- a/CMake/dcmqiVersionConfigure.h.in
+++ b/CMake/dcmqiVersionConfigure.h.in
@@ -15,5 +15,8 @@
 
 #define dcmqi_WC_REVISION "@dcmqi_WC_REVISION@"
 #define dcmqi_WC_URL "@dcmqi_WC_URL@"
+#define dcmqi_WC_TAG "@dcmqi_WC_TAG@"
+
+#define dcmqi_INFO "dcmqi revision hash: @dcmqi_WC_REVISION@ tag: @dcmqi_WC_TAG@"
 
 #endif // __dcmqiVersionConfigure_h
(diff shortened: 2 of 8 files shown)


jcfr:
> This commit adds console output of the git hash and tag

Adding version info to the converter makes sense :+1: 


> especially with the use of Docker images, it is not easy
to determine the version of the software being used.

That said,  following https://github.com/QIICR/dcmqi/commit/3d01c021dcce4c770f901f0b8f4ee3652a8ecb69 build metadata are included in the docker image.

This means that they can be retrieved from the image doing the following:

```bash
$ docker inspect --format "{{ index .Config.Labels }}" qiicr/dcmqi:v1.0.1
map[license:GPLv2 org.label-schema.build-date:2017-02-20T20:52:28Z org.label-schema.vcs-url:git@github.com:QIICR/dcmqi.git build-date:2016-03-31 name:CentOS Base Image org.label-schema.name:qiicr/dcmqi org.label-schema.schema-version:1.0 org.label-schema.vcs-ref:7f40682 vendor:CentOS]
```

```bash
$ docker inspect --format "{{ index .Config.Labels \"org.label-schema.vcs-url\" }}" qiicr/dcmqi:v1.0.1
git@github.com:QIICR/dcmqi.git

$docker inspect --format "{{ index .Config.Labels \"org.label-schema.vcs-ref\" }}" qiicr/dcmqi:v1.0.1
7f40682
```

fedorov:
> Adding version info to the converter makes sense

It would be quite helpful if something like this was supported automatically by SlicerExecutionModel! ;-)

> build metadata are included in the docker image

Sorry, I didn't do a thorough enough search to figure out how to get that metadata. I find it extremely confusing that it is not exposed in the Docker Hub interface! This is indeed very helpful, and complementary to what I added in this commit, thank you for adding this note. Having the version printed out on invocation of the converter is much easier for the user to send back to the developers. 

> `license:GPLv2`

Do you know where this is coming from and how to fix it? This bothers me quite a lot, since this can automatically turn off commercial users.

fedorov:
@jcfr is the image labeled with GPLv2 because its base image is CentOS, which is under that license?

I thought so after looking at the info here: https://microbadger.com/images/qiicr/dcmqi

> LABEL name=CentOS Base Image vendor=CentOS license=GPLv2 build-date=2016-03-31

jcfr:
> Sorry, I didn't do a thorough enough search to figure out how to get that metadata. I find it extremely confusing that it is not exposed in the Docker Hub interface!

But it is exposed from the microbadger link of the README: https://microbadger.com/images/qiicr/dcmqi

> the version printed out on invocation of the converter is much easier for the user to send back to the developers.

:+1:  

Just thought it would be relevant to mention the docker metadata

> Do you know where this is coming from and how to fix it? 

This is the license associated with centos, or more generally any linux distribution.

Some reference I could find ... 

https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers/#comment-1682
https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.en.html#MereAggregation

fedorov:
> https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers/#comment-1682

Reading the reference above, to me it means that one cannot distribute tools compiled from BSD licensed source code in a docker image that has a layer covered by GPL, is this your interpretation as well?

@pieper did you think about this, any comments?

jcfr:
Other comment on this topic:  http://opensource.stackexchange.com/a/4689

Look like this is not an issue. That said checking with layer may be best ...

pieper:
The licensing question is interesting.  The typical linux kernel exemption
and the side-by-side distribution criteria do make the GPL seem
non-applicable to the docker image that is used like a shell or desktop.
But when you bundle a proprietary CLI inside of an image that relies on GPL
libraries to run, then I could see this being closer to the exact situation
the GPL tries to prevent.

In our case I suspect we could work around it because I believe it is
possible to allow the code to be dual-licensed under BSD (or Slicer
license) and GPL letting the user decide which terms they want to follow.
That would be ugly, but possible as a fallback.

PR #208, opened 2017-02-23
ENH: webapp for PM 

### LINKED ISSUE #202: Add web app for PM converter
@che85 now that we have PM context in place, it would be great to add this webapp. Can you walk me over the steps and tools you use to do this? Maybe we can plan to work on this together on Tuesday?

TODO:

- [x] https://github.com/QIICR/dcmqi/blob/master/doc/schemas/pm-schema.json#L75 needs to be a number

che85:
fixes #202 

for testing purposes: http://che85.github.io/dcmqi/#/pmap

TODO:
- [x]  update the pm contexts to correctly list units in lists
- [x] reset form properly
- [x] check DICOM standard for required attributes
- [x] fix bugs that were invented by generalizing (seg converter)


fedorov:
* let's separate units from quantity, and put quantity before anatomic location in the form layout
* units are actually defined in the schema as arrays of items, not sure why validators didn't catch this for the schema instantiations: https://github.com/QIICR/dcmqi/blob/master/doc/schemas/pm-context-schema.json#L30. Let's automatically select the first code in the list of units codes.

fedorov:
and I should also update the pm contexts to correctly list units in lists

che85:
@fedorov let me know if you have any comments

fedorov:
This looks great, just series description should be initialized to "Parametric map", or "<quantity code> parametric map".

Have you tested this with the updated PM contexts from #215? It has also occurred to me that we could fix the problem of initializing DerivedPixelContrast for #216, since it is essentially defined by the quantity, and I should also add definition of the quantity to the schema. So let me address those issues first, and then merge #215, and then get back to this PR, ok?

che85:
@fedorov renamed schema defaults in #218

will test #215 and get back to you

che85:
#215 works for me

image

what I don't like is the spacing. I am going to do something about that.

fedorov:
@che85 can you please add the code to propagate the value of DerivedPixelContrast from the contexts update in https://github.com/QIICR/dcmqi/pull/215/commits/9e5dd870d20e14cb4f626578e63e5c68539fa5e4 into the web app generated json?

fedorov:
> can you please add the code to propagate the value of DerivedPixelContrast from the contexts 

as discussed further in #216, this will be simplified by #219, and so please ignore this request!

fedorov:
@che85 I assume this one is not ready to test/merge until you tell me.

che85:
@fedorov let's merge it. I can always add new functionality

fedorov:
let me quickly make sure nothing stands out, since I want to use it very soon for a PM tutorial

fedorov:
Can you remove DerivationCode from the list of required attributes here: https://github.com/QIICR/dcmqi/blob/master/doc/schemas/pm-schema.json#L10?

I don't recall why we added it there. I do not think it is required. Otherwise we have the error below, which only adds confusion to the user who I think would expect valid output from this generator.

image

che85:
@fedorov please try again. I removed DerivationCode as required in PR #221

PR #209, opened 2017-02-23
BUG: MeasurementUnitsCode attribute should be a list

fedorov:
Somehow this error is not detected by ajv. We should report it.

Re #208


che85:
Let me check our code first....

fedorov:
> Let me check our code first....

Sure .. but just to clarify - what code?

che85:
look at our tests and or schemas

fedorov:
makes sense. I ran validation with and without the lists, and it always validates.

PR #211, opened 2017-02-23
ENH: updating DICOM segmentation context

fedorov:
Comments from @dclunie:

I have got to the stage where I am able to generate a JSON
file that should match the schema that you are using in
dcmqi, with some changes, and can do so using the source
of the standard (modified by various in progress CPs), so
that in future we can fully automate updates.

Please let me know if:

(a) this works with your application (I am not sure how
to fire it up or I would check myself)

(b) the contents seem reasonable/sufficient

One of the github comments suggested elimination of duplicates,
so I have done that, but still tracked the (multiple) CIDs
used, and I added a few extra attributes to the JSON
to track a few things that you probably don't need.

This actually turned out to be more interesting than I
had expected, since resolving differences between what is
currently in the standard, what is in the Slicer LUT
spreadsheet, and determining what is needed to determine
pairedness (and hence the need for a laterality modifier)
in an automated manner required some creative use of
stylesheets and coordination of multiple sources (the
DICOM standard for the property category and type context
groups, UMLS to establish a mapping to FMA, some FMA
post-processing to determine pairedness on an ontological
basis and then some sorting to detect and merge duplicates.

I also checked for inconsistencies automatically and
resolved most of them, either by extending the DICOM CPs,
or changing (my local copy of) the Slicer LUT spreadsheet
(vide infra).

Anyhow in this DropBox folder:

https://www.dropbox.com/sh/lam877z0avflt2j/AAANgOIYWTwNaUtl1A4EsV7Pa

I have uploaded:

  jsonoutputfile_20170210.json.zip

which contains the JSON terminology

  cp1694_02_matchsegmentcategorieswithcids.pdf

which contains the necessary changes to DICOM

   NOTES_ChangesToSlicerLUTSpreadsheetRequired.txt

which describes the changes I needed to the spreadsheet

[...]

Note that I have not yet attempted to reconcile the category
choices in the spreadsheet with the category choices that
result from using cp1694 as the source, which might be
interesting.

I have not messed with any of the github tracker items, since
there seemed to be quite a few related to this subject and I
was not sure which are active or not.

PS. Note that I did not use the ClearCanvas AIM XML files again;
I did this from first principles using the DICOM standard (as
amended by the WIP CP 1694). I will regenerate new ClearCanvas
AIM XML config files to use with that tool later if necessary.

Clarification on the addition of the "paired" attribute:

The "paired" attribute describes whether the anatomy concept:

- "N" is already lateralized, e.g., left hand

- "Y" is paired (can have laterality), e.g., hand

- "U" is unpaired (cannot have laterality), e.g., spine

I intend to document this using CP 1695 (which is a work in
progress).

Figuring out whether or not a concept is paired requires
consulting an ontology ... at the moment I use the FMA
OWL file as the source for this and trawl through it looking
for concepts that have a child with "left" or "right" in
their label, and then walks back up the class hierarchy
propagating the pairedness to parent classes. This is crude
but mostly works. I will be correlating this with whatever
laterality information SNOMED contains when I figure that
out, as well as performing manual sanity checking.

The "paired" attribute controls whether or not the laterality
modifiers are added to the JSON file (i.e., only if "Y"),
and it doesn't really need to be in the JSON file itself,
except that it seemed reasonable to add it ... for that
matter, I was not sure which of the existing attributes your
code actually used.

The placeholder CIDs are there until CP 1694 reaches final
text.


fedorov:
@che85 I was trying to test this in the web app locally, and changed just one line to replace the Slicer seg context URL with https://raw.githubusercontent.com/fedorov/dcmqi/216a7f290bcc942e0e3ec6af13408965c17bb412/doc/segContexts/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json, but I am getting this error

image

Maybe you can quickly test if this json works with the web app?

che85:
Yeah let me check

che85:
@fedorov works for me...

fedorov:
ok, that is good, thanks

che85:
Are you still working on this or is it ready for merge?

fedorov:
No, not ready to merge, I was checking things, and now emailed for a clarification to David. I will rename it.

fedorov:
How about we follow the convention that as long as prefix is 'WIP' it means 'not ready for merge'?

che85:
ok. I just wanted to check because #214 validation depends on this PR

fedorov:
Clarification re alphanumerical CIDs:

> > David, what is the meaning of the suffix you append to "cid" attribute
> > in some cases - like "6040rf3" ? Wanted to confirm this is not a
> > mistake.
> >
> No, these are new context groups caused by refactoring to
match the categories, per CP 1694.

PR #214, opened 2017-02-24
ENH: schema correction, misspelling, typos

che85:
* added additional properties=false to detect missing schema definitions
* TODO:
  - [x] not sure about the `comment` field which probably doesn't belong into the schema but validation fails if not
  - [x] rgb values in segmentation context files need to be converted to integer
  - [x] 3dSlicerLabel is not defined for "Modifier"

depends on #211


doc/schemas/pm-context-schema.json
```diff
@@ -21,13 +22,16 @@
         "contextGroupName": {
           "type": "string"
         },
+        "comment": {
+          "type": "string"
+        },
```
che85: not sure about that field. I added it because there are two comments in the context files and validation fails if the definition here doesn't exist

doc/schemas/pm-context-schema.json
```diff
@@ -21,13 +22,16 @@
...
         "UMLConceptUID": {
           "type": "string"
         },
         "SNOMEDConceptID": {
           "type": "string"
         },
-        "MeasurementUnitsCodes": {
+        "MeasurementUnitsCode": {
```
che85: @fedorov That is why the validator didn't throw errors

che85: P.S. setting `"additionalProperties": false` is recommended for objects

doc/schemas/pm-context-schema.json
fedorov: I added those more as notes to myself, but perhaps it makes sense to keep it. The issue is that there is inconsistency in how code meanings are defined (i.e., sometime code meaning is an abbreviation, such as IAUC, and sometimes it is fully expanded definition, such as "Regional Cerebral Blood Flow"). It is confusing, and we may want to include some clarification in the comment.

doc/schemas/pm-context-schema.json
fedorov: I think you already told me about this, but I forgot. Glad to know it is not the issue of validator!

che85:
@fedorov once #211 is merged, this PR should validate without errors

che85:
@fedorov ready to merge if you agree

fedorov:
> 	`Merge branch 'master' into fixed_schemas`

@che85 you should consider using git rebase next time instead of the merge.

che85:
yeah I know. I wanted to give it a try ..

PR #215, opened 2017-02-24
ENH: update diffusion and perfusion parametric map contexts

fedorov:
These are derived automatically from the XML of the standard and/or correction
proposals that establish explicit connection between the quantities and quantity
units. CP-1665 describes the modifications to the standard related to the diffuson
context. CP with the modifications to the perfusion quantities is not yet assigned
a number.

Note from @dclunie:

> The choice to use an abbreviation (Ktrans, T1) or a fully spelled
> out Code Meaning ("Regional Cerebral Blood Flow") is historical
> and arbitrary, except in that "RCBF" would be very confusing
> (relative versus regional question).

This means that we might consider including definition of the quantity in the
context, since code meaning alone can be too confusing.


fedorov:
from a conversation with @dclunie below, so let's leave definitions out of the context, for now:

```
On 2/24/17 12:20 PM, Andrey Fedorov wrote:
>> The choice to use an abbreviation (Ktrans, T1) or a fully spelled
>> out Code Meaning ("Regional Cerebral Blood Flow") is historical
>> and arbitrary, except in that "RCBF" would be very confusing
>> (relative versus regional question).
>>
>
> Oh boy. In light of this, I think we should include definition in the
> quantity, since code meaning alone is clearly not sufficient. What do
> you think?

Well, I think there is a practical limit here. Most things
have sufficient meaning in context and there is no point in
reproducing the entire standard in the user interface.

We could probably add a URL to hyperlink to the standard
definition though (once they are final text and included in
a release), so that a dligient user could follow them.

These are of a predictable form, e.g.:

http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_110809
```

fedorov:
Regarding the failing test `ajv_pm-example`, I wonder if this is because the attribute validates against both of the schemas in `anyOf`. According to http://json-schema.org/example2.html

> the oneOf keyword is new in draft v4; its value is an array of schemas, and an instance is valid if and only if it is valid against exactly one of these schemas

So perhaps this attribute should be just CS.

fedorov:
I think this one is ready for review/merge

PR #221, opened 2017-03-01
ENH: removed DerivationCode as required attribute

che85:


che85:
@fedorov I am merging this

fedorov:
Ok, here are some bad news!

1. Derivation Image is required! :-( http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.75.5.html

2. Running the converter fails when I do it locally with `DerivationCode must be specified in the input metadata!`

3. Tests on CircleCI are passing!

So I guess we need to (in the order of priorities!):

1. Figure out why tests are passing!

2. make DerivationCode required again

3. Initialize it in the web app (I suggest to set it to the same value as the quantity code until we come up with a better idea). Let's discuss with David tomorrow.

che85:
That's how you implemented it in https://github.com/QIICR/dcmqi/blame/338fd2317959b3270168824797f71ff24a6ced7c/libsrc/ParaMapConverter.cpp#L308

I will check why CircleCI tests are passing...

fedorov:
oh, it's totally my fault - it is definitely required, according to the standard.

che85:
So I will put it back then and assign default values in the schema?

fedorov:
Ok, I understand why tests do not fail. We still have those attributes in the input json for examples. 

> So I will put it back then and assign default values in the schema?

Yes. Sorry for the confusion I created. Let's use the quantity code for the derivation code.

Issue #192, opened 2017-02-01
Mapping of the ITK image slices to DICOM instances should be factored out of SEG converter


fedorov:
Function in question: https://github.com/QIICR/dcmqi/blob/master/include/dcmqi/ImageSEGConverter.h#L51

It is also needed by the PM converter

To be discussed in the context of refactoring of the code organization.

Issue #150, opened 2016-11-18
Support conversion of segmentations of multi-frame input series


fedorov:

che85:
That's after RSNA, right?

fedorov:
Absolutely! :)

I was just going over the code trying to understand why Mint people have segfault (they do not have a complete example to reproduce) and saw that comment.

che85:
Well then they should try to provide a complete example for us. Otherwise It's really hard to find the reason for its failing....

fedorov:
yes, that's what I asked. Our test datasets work fine for them.

che85:
Did they ever run dciodvfy on their data? Maybe there is something wrong with their input data?

fedorov:
Same requirement applies to PM

PR #226, opened 2017-03-10
WIP: refactoring of the multiframe conversion implementation

fedorov:
**WIP == FYI == "do not even consider merging"**

This is an intermediate commit that aims to improve organization of the conversion
functionality, motivated by the various standing issues (e.g., #217, #192 and #150)
that are difficult to address with the current level of disarray in the code.

Some of the ideas are the following:

 - introduce basic hierarchy of classes to allow reuse of the common
   functionality between PM and SEG converters (MultiFrameConverter parent for
   ParametricMapConverter and SegmentationImageConverter)
 - do not use static functions, instead make better use of converter classes
 - factor out containers to keep information about ImageVolume and DICOMFrame
   to simplify conversion and book keeping
 - improve consistency of naming and code style

My plan to proceed with the refactoring is to implement "shadow" functions in the
updated classes while keeping the original functions initially, and then phase out
the original functions.

The current commit is just an intermediate milestone to open this for feedback
and keep development history at least somewhat incremental.


include/dcmqi/DICOMFrame.h
```diff
@@ -0,0 +1,36 @@
...
+  public:
+    // distinguish between the frames from the legacy data and enhanced multiframe objects
+    enum {
+      LegacyInstanceFrame = 0,
+      EnhancedInstanceFrame
+    };
+
+    DICOMFrame(DcmDataset *dataset, int number) :
+        frameNumber(number),
+        frameDataset(dataset) {
```
che85: indentation ..

include/dcmqi/MultiframeConverter.h
```diff
@@ -35,15 +35,21 @@
...
 
+// common type definitions
 typedef short ShortPixelType;
 typedef itk::Image<ShortPixelType, 3> ShortImageType;
 typedef itk::ImageFileReader<ShortImageType> ShortReaderType;
 
 namespace dcmqi {
 
-  class ConverterBase {
+  class MultiframeConverter {
```
che85: MultiframeConverter should be MultiFrameConverter


ifdefs should be renamed according to class name
```
#ifndef DCMQI_CONVERTERBASE_H
#define DCMQI_CONVERTERBASE_H
```

include/dcmqi/MultiframeObject.h
```diff
@@ -0,0 +1,136 @@
...
+#include "dcmqi/ImageVolumeGeometry.h"
+
+using namespace std;
+/*
+ * Class to support conversion between DICOM and ITK representations
+ * of multiframe objects
+ *
+ *
+ */
+class MultiframeObject {
```
che85: MultiframeObject --> MultiFrameObject

include/dcmqi/MultiframeConverter.h
pieper: The filenames use "Multiframe" but some of the code uses "MultiFrame" so it should be uniform in the codebase.

The DICOM standard uses "Multi-frame" which can't be used as an identifier, so that doesn't help.

It looks like DCMTK has a mix of MultiFrame and Multiframe.  

Personally I prefer Multiframe and would suggest using that consistently and that looks like Andrey's preference too.

pieper: @michaelonken is there a preference in DCMTK for Multiframe vs. MultiFrame?

I see several examples of both currently...

(cd DCMTK; git grep -i multiframe) | less

fedorov: I did pay attention to this, and my logic is to capitalize individual words. Since they are not separate words in the standard, I decided not to capitalize Frame.

pieper:
Your English instincts are quite good Andrey.  In general English usage if
a hyphenated word gets a lot of usage the hyphen eventually gets dropped so
let's do go with Multiframe everywhere.

libsrc/ParametricMapObject.cpp
```diff
@@ -0,0 +1,230 @@
...
+  initializeVolumeGeometry();
+
+  // NB: the sequence of steps initializing different components of the object parallels that
+  //  in the original converter function. It probably makes sense to revisit the sequence
+  //  of these steps. It does not necessarily need to happen in this order.
+  initializeEquipmentInfo();
+  initializeContentIdentification();
+
+  // TODO: consider creating parametric map object after all FGs are initialized instead
+  createParametricMap();
```
che85: I think the method name should indicate the type (DICOM or ITK):

We could add two virtual methods {create/get}{DICOM/ITK}Representation to MultiframeObject that needs to be implemented by inheriting classes.

fedorov:
is expected to resolve #192

PR #227, opened 2017-03-14
BUG: fix referenced instances item initialization

fedorov:
Add referenced series item only if it is not empty.


fedorov:
Reported by a user!

PR #231, opened 2017-03-30
BUG: fixes Bad XML for segimage2itkimage (issue #230)

### LINKED ISSUE #230: Bad XML for segimage2itkimage
The XML produced by segimage2itkimage specifies `outputDirName` with a `channel` of `input` (see the snippet below).  However, as the name suggests, the `channel` should be `output`.  I doubt that this will work inside Slicer.

Can this be fixed?

Thanks,
-dan

```
    <file>
      <name>outputDirName</name>
      <label>Output directory name</label>
      <channel>input</channel>
      <longflag>outputDirectory</longflag>
      <description>Directory to store individual segments saved using the output format specified files. When specified, file names will contain prefix, followed by the segment number.</description>
    </file>
```

che85:
fixes #230

### DIFF
--- a/apps/seg/segimage2itkimage.xml
+++ b/apps/seg/segimage2itkimage.xml
@@ -22,7 +22,7 @@
     <directory>
       <name>outputDirName</name>
       <label>Output directory name</label>
-      <channel>input</channel>
+      <channel>output</channel>
       <longflag>outputDirectory</longflag>
       <description>Directory to store individual segments saved using the output format specified files. When specified, file names will contain prefix, followed by the segment number.</description>
     </directory>


fedorov:
@blezek, I am merging this, since it can't hurt. Whenever I get a chance I will discuss with @pieper or @jcfr to clarify semantics of SEM.

I won't be able to generate packages until next week though, since for now I am generating and uploading them manually, since our automatic CI packaging is not yet working.

fedorov:
@che85 can you please review other CLIs and make sure the assignment of channel type is consistent in other places?

che85:
@fedorov I checked once already, but I can check again to make sure.

fedorov:
Thanks Christian. you didn't mention explicitly initially, so I was not sure.

che85:
Not sure about that

https://github.com/QIICR/dcmqi/blob/master/apps/paramaps/paramap2itkimage.xml#L34-L57

there is no channel defined at all.

che85:
@fedorov All other apps are correct

pieper:
@fedorov I think you heard from the real authority because Dan Blezek was one of the original developers of the CLI XML (along with Jim Miller and Bill Lorensen) in the early days of NA-MIC.

blezek:
But I'm old and grey now and forgot all the things I did in my youthful years... :smile:

`channels` are only needed when this is I/O involved, and are clues to the consumer of the CLI as to how to get data in and out of the plugin.  Frankly, I'm surprised the CLI stuck around so long.

fedorov:
Ok, thank you Dan and Steve! Let me know if there are other instances where
we need to update XML syntax. If I don't forget, I will add this
clarification to the wiki (am on the plane right now writing emails in
offline mode).

PR #242, opened 2017-04-26
BUG: after SegmentedPropertyType selection Modifier selector got neve…

che85:
…r enabled

available for testing on http://che85.github.io/dcmqi/#/seg once GitHub cache gets updated

### DIFF
--- a/js/controllers.js
+++ b/js/controllers.js
@@ -814,6 +814,7 @@ define(['ajv'], function (Ajv) {
       }
       $rootScope.$emit(self.selectionChangedEvent, {
         item:self.selectedItem,
+        segmentNumber:$scope.segmentNumber,
         value:item ? item.object : item,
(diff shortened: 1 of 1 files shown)


js/controllers.js
```diff
@@ -814,6 +814,7 @@ define(['ajv'], function (Ajv) {
       }
       $rootScope.$emit(self.selectionChangedEvent, {
         item:self.selectedItem,
+        segmentNumber:$scope.segmentNumber,
```
fedorov: It appears to work, but I have to say that by simply looking at the variables names and context - looks like magic to me!

che85: There is a segmentNumber comparison so that the current segment (displayed tab segment number) matches the segmentNumber for the selector so that not accidentally another segments attribute get changed (emitted)

PR #254, opened 2017-05-19
ENH: adding checks for existence of file/files in file list (issue #225)

### LINKED ISSUE #225: Check whether input files/output paths exist
Tell users if they don't exist to help troubleshoot problems. Print absolute path for the input filenames/directories passed by the user.

- [x] itkimage2paramap
- [x] paramap2itkimage
- [x] itkimage2segimage
- [x] segimage2itkimage
- [x] tid1500reader
- [x] tid1500writer

che85:
* cerr upon missing file/directory and EXIT_FAILURE

fixes #225

### DIFF
--- a/libsrc/Helper.cpp
+++ b/libsrc/Helper.cpp
@@ -4,6 +4,34 @@
 
 namespace dcmqi {
 
+  bool Helper::isUndefinedOrPathDoesNotExist(const string &var, const string &humanReadableName) {
+    return Helper::isUndefined(var, humanReadableName) || !Helper::pathExists(var);
+  }
+
+  bool Helper::isUndefinedOrPathsDoNotExist(vector<string> &var, const string &humanReadableName) {
+    return Helper::isUndefined(var, humanReadableName) || !Helper::pathsExist(var);
+  }
+
+  bool Helper::pathsExist(const vector<string> &paths) {
+    bool allExist = true;
+    for(vector<string>::const_iterator iter = paths.begin(); iter != paths.end(); ++iter) {
+      if (!Helper::pathExists(*iter)){
+        allExist = false;
+      }
+    }
+    return allExist;
+  }
+
+  bool Helper::pathExists(const string &path) {
+    struct stat buffer;
+    if (stat (path.c_str(), &buffer) != 0) {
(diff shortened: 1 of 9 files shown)


che85:
Maybe also should add some ctests where it is expected to fail

fedorov:
Looks good!

The only issue I have with this PR is that the naming of the functions is a bit confusing `exists`, `filesExist`, `isEmptyOrDoesNotExist`, `isEmptyOrFilesDoNotExist`. 

It is not always clear (and maybe inconsistent) what `exist` refers to. It looks like sometime "isEmpty" checks whether a vector is empty, sometime it checks whether directory is empty, but it can also check whether file exists ... Since `dcmqi::helper` is not a file-system-operations-specific helper, it probably makes sense to include explicitly what is being checked into the function name. What do you think?

I was also confused by this line: `helper::isEmpty(outputParaMapFileName, "Output DICOM file"))`. What is the expected behavior here? What would it mean for an output file to be empty?!

How about the following names:
* `pathExists`: both for directories and individual files
* `pathsExist`: for checking lists of paths
* I am not sure what `isEmpty` as applied to an output file should be. I guess you wanted to have a single function to check the length of a string and the length of an array? I am NOT sure it is a good idea! I would just use string::size() explicitly when string size needs to be checked.

che85:
@fedorov Please take a look at the method names now. It should be more readable now. Let me know if you have any objections

libsrc/Helper.cpp
```diff
@@ -49,7 +77,7 @@ namespace dcmqi {
           dcmDatasets[i]->findAndGetOFString(DCM_SOPInstanceUID, tmp);
           if (tmp == sopInstanceUID) {
             cout << dicomImageFiles[dcmFileNumber].c_str() << " with SOPInstanceUID: " << sopInstanceUID
-                 << " already exists" << endl;
+                 << " already pathExists" << endl;
```
fedorov: was this replaced accidentally?

che85: Oh yeah! Accidentally....

che85:
@fedorov I squashed the commits. Feel free to merge.

Issue #255, opened 2017-05-20
Build is failing on CircleCI


fedorov:
@jcfr something changed recently, and CircleCI started to fail in this step:

```
# Install tools required to run DCMQI "doc" tests
cd /home/ubuntu/dcmqi/docker/.. && \
	/tmp/dockcross bash -c "export PATH=/work/build/node-v6.9.5-linux-x64/bin:$PATH && npm install jsonlint -g && npm install ajv-cli -g"
npm ERR! Linux 3.13.0-117-generic
npm ERR! argv "/work/build/node-v6.9.5-linux-x64/bin/node" "/work/build/node-v6.9.5-linux-x64/bin/npm" "install" "jsonlint" "-g"
npm ERR! node v6.9.5
npm ERR! npm  v3.10.10
npm ERR! path /home/ubuntu/.npm
npm ERR! code EACCES
npm ERR! errno -13
npm ERR! syscall mkdir

npm ERR! Error: EACCES: permission denied, mkdir '/home/ubuntu/.npm'
npm ERR!     at Error (native)
npm ERR!  { Error: EACCES: permission denied, mkdir '/home/ubuntu/.npm'
npm ERR!     at Error (native)
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'mkdir',
npm ERR!   path: '/home/ubuntu/.npm' }
npm ERR!
npm ERR! Please try running this command again as root/Administrator.

npm ERR! Please include the following file with any support request:
npm ERR!     /work/npm-debug.log
make: *** [prereq.install_npm_packages] Error 243
```
Running `npm install jsonlint -g` without `dockcross` works.

I tried [with](https://circleci.com/gh/QIICR/dcmqi/648) and [without](https://circleci.com/gh/QIICR/dcmqi/649) cache.

Do you know what is going on?

fedorov:
If I run this line with sudo: `/tmp/dockcross bash -c "export PATH=/work/build/node-v6.9.5-linux-x64/bin:$PATH && npm install jsonlint -g && npm install ajv-cli -g"`, I get past that error, but then I get another error:

```
[...]

PR #256, opened 2017-05-20
BUG: run npm with sudo to fix the issue on CircleCI

fedorov:
see #255


fedorov:
@jcfr looks like this fixed #255, but please follow up in case this should be addressed differently.

PR #267, opened 2017-06-23
ENH: updated dcmtk revision, included header files in the cmake library sources

che85:
* added includes in CMakeLists.txt for allowing CLion to resolve
references


che85:
👍

fedorov:
There goes the mystery of CircleCI untriggered build, I hope! :-)

fedorov:
@che85 so it was good to wait for CircleCI results - apparently the issue was not caught by the other two, although I don't know why.

PR #269, opened 2017-06-28
Make composite context/image library directories parameters optional in tid1500writer

fedorov:


fedorov:
@nolden @JasminMetzger this issue we discussed today should now be resolved and I think absolute paths should work for the metadata JSON passed to `tid1500writer`.

PR #270, opened 2017-06-29
ENH: add a test of converting pm with no derivation items

fedorov:
NB: the new test is expected to fail due to known problems leading to segfault
that will be fixed in another commit


fedorov:
This new test exposed another problem, or perhaps multiple problems!

```
10: The size of the Baseline image and Test image do not match!
10: Baseline image: /Users/fedorov/github/dcmqi/data/paramaps//pm-example-3slices.nrrd 
  has size [255, 252, 3, 1, 1, 1]
10: Test image:     /Users/fedorov/local/builds/dcmqi-refactored/dcmqi-build/Testing/Temporary/pmap/makeNRRDParametricMap-3slices-pmap.nrrd 
  has size [252, 255, 3, 1, 1, 1]
```

By accident, I created a sample dataset that has a different matrix size (I was using CropVolume from Slicer), and this exposed a problem with change in the dimensions in the output.

Furthermore, the resulting image from the round trip is completely messed up, perhaps because now the matrix size is "weird" (252x255) and something is done differently somewhere!

image

Not at all as simple as I was hoping!...

fedorov:
Tests are passing locally. Confirmed that visually the result also makes sense now.

image

fedorov:
> NB: the new test is expected to fail due to known problems leading to segfault

all tests are expected to pass now, PR rebased to the current master

Issue #253, opened 2017-05-19
Add segmentation file name optional attribute to itkseg2segimage JSON


fedorov:
Following up on #240, it will help integration with Mayo Grunt, and it might help make the usage easier, if the individual sections corresponding to the segments in JSON could be associated explicitly with the segmentation files passed on input.

The idea we agreed upon is to add an optional attribute to each segment group (`segmentationFileName`?) that would correspond to the filename (without path qualifiers) passed to `--inputImageList`.

fedorov:
@blezek @potis sorry for the huge delay, there is always something, but I finally got to work on this. I think #278 should address this need, but can you take a quick look and confirm before we merge?

What I did is I added a new optional attribute to the segmentation parameters schema called `segmentAttributesFileMapping`. When this attribute is present, it should contain as many items as the `segmentAttributes` list, and will contain strings that should uniquely map the items of the `segmentAttributes` list to the file names passed in the command line. The code will search each of the input ITK image file names for exact match of the items in this list, and match ITK images to the items in `segmentAttributes` accordingly. Can you confirm if this will address the issue related to explicitly specifying the order of files that we discussed in #240? Thanks

fedorov:
@blezek @potis - any updates on the grunt side?

fedorov:
@blezek @potis - grunt came up in a discussion today - please let me know if you have any updates/feedback

blezek:
@fedorov @potis  I poked around this morning.  Grunt didn't recognized this bit of XML:

```
   <string-vector>
      <name>dicomImageFiles</name>
      <label>DICOM images file names</label>
      <channel>input</channel>
      <longflag>inputDICOMList</longflag>
      <description>Comma-separated list of DICOM images that correspond to the original image that was segmented. This means you must have access to the original data in DICOM in order to use the converter (at least for now).</description>
    </string-vector>
```

We are going to have to look into what it might take to support, or write a shell script to support.  I'll check with @potis this week.

fedorov:
@blezek can you update the status?

fedorov:
@blezek @potis any hope for this issue?

Looking at https://github.com/Mayo-QIN/grunt, last commit was in Oct 2017. Is that repo maintained/used by anyone?

PR #278, opened 2017-07-17
ENH: adding attribute to link segmentAttributes sub-arrays to specific files

fedorov:
Compiles, need to test

Aims to address #253


include/dcmqi/itkTestMain.h
```diff
@@ -89,7 +89,7 @@ int main(int ac, char *av[])
 {
   itk::FloatingPointExceptions::Enable();
 
-  double       intensityTolerance  = 2.0;
+  double       intensityTolerance  = 0.0001;
```
fedorov: @che85 this was a nasty bug!

fedorov:
@che85 if CIs are green and you have no edits requested, hold on merging for a little bit at least until we hear from @blezek/@potis.

include/dcmqi/itkTestMain.h
che85: That code came from BrainTools.... yeah it's a pretty high tolerance ...

fedorov: Especially for comparing segmentations - they can be very different pixel-wise, but test would still succeed!

che85: Glad that you noticed it....

che85:
LGTM

blezek:
@potis and I reviewed it today, looks good, but we don't have test data.  Could you suggest a test so we can verify that the Docker works as expected?

Go ahead and merge when you are ready.

fedorov:
@blezek we have a test for this, see

https://github.com/QIICR/dcmqi/blob/master/apps/seg/Testing/CMakeLists.txt#L43 (itk2dcm)

and

https://github.com/QIICR/dcmqi/blob/master/apps/seg/Testing/CMakeLists.txt#L127 (dcm2itk for the output of the first one)

Is this what you were looking for?

PR #282, opened 2017-07-25
BUG: fix the order of parameters to import() call

fedorov:
Due to a fixed bug in DCMTK - see discussion in

https://github.com/QIICR/QuantitativeReporting/issues/170#issuecomment-317540337

the order of parameters passed to import changed and dcmqi was incorrectly
importing series attributes, and not importing frame of reference.


fedorov:
@che85 since you are working on #226, can you fix it in that branch too?

PR #298, opened 2017-08-29
Fix setting of cxx standard

jcfr:


fedorov:
@nolden do you want to test if this is working for you before we merge it, since you contributed the earlier patch related to C++11?

jcfr:
@nolden Any objection moving forward with this change ?

nolden:
Looks good to me, thanks for cleaning it up!

PR #299, opened 2017-08-29
Fix warnings

jcfr:


fedorov:
Thank you @jcfr! Sorry we have not fixed this earlier.

PR #304, opened 2017-09-01
Fix some CXX11 errors

nolden:
- Pass C++ standard version to ITK in superbuild
- Use DCMTK iterator in the correct way, to make it work with a DCMTK that is configured to use STL


fedorov:
Marco, do you have any concerns about #298? Should we go ahead with merging it?

Issue #305, opened 2017-09-07
Normative stats in DICOM-SR .json


EmilyLindemer:
Hi @fedorov, this issue is in regards to our recent meeting about normative statistics for neuroimaging applications. 

I was using [this](https://github.com/QIICR/dcmqi/blob/master/doc/examples/sr-tid1500-ct-liver-example.json) template to create my own json file to create DICOM-SR's, but doesn't cover my specific use case. 

Lines 134-186 of the template linked above show how to include structural statistics such as min, max, and stddev for a given structure _at the individual level._  We would like to include individual-based statistics _as well as_ group-based statistics. 

For example, if we are creating a DICOM-SR that reports regional brain cortical thickness values, then for each structure we would want to include the following: 

1. Minimum thickness of the structure at the individual level
2. Maximum thickness at the individual level
3. Mean thickness at the individual level
4. Std dev of thickness across the individual's structure
5. Minimum thickness of the structure based on a large group of age-matched individuals
6. Maximum thickness based on group
7. Mean based on group
8. Std dev based on group
9. Z-score of individual's thickness compared to the group mean distribution for thickness

Importantly, we want to ensure that the values are interpreted properly by a clinician. In other words, make sure that they don't think that the value reported for (7) above is the individual's mean thickness. 

The paper attached below demonstrates how we plan to derive normative statistics for each age range and brain structure, and it is these values that we are looking to include in our DICOM-SR along with individual-level statistics. 

Thank you! 

[Potvin2016.pdf](https://github.com/QIICR/dcmqi/files/1285886/Potvin2016.pdf)

fedorov:
@EmilyLindemer excellent - thank you for this summary! We will discuss with @dclunie and follow up in this issue.

fedorov:
@EmilyLindemer we did discuss this with @dclunie, and I forwarded him this question. I will use this issue to continue the discussion and provide updates.

One question that came up is what do you mean by "based on a large group of age-matched individuals"? 

Do you mean that 

1. you would compare to an existing normative group defined in a reference, such as the one you included from Potvin et al., 

or 

2. did you mean you intend to segment a large group of individuals, and then report minimum thickness over the cohort of those individuals?

(1) is suitable for DICOM encoding, while (2) is a cohort-based measurement, and would not fit into the existing DICOM model (if I understood correctly the various discussions we had with David before).

Please let us know.

EmilyLindemer:
[...]

PR #309, opened 2017-11-04
WIP: add support for initializing measurement properties

fedorov:
This aims to address #305. Schema extended to support numeric and text properties
of the measurement, as suggested by @dclunie in

https://github.com/QIICR/dcmqi/issues/305#issuecomment-335172718

Since these attributes are not supported by the dcmsr wrapper class, they are added
in a separate iteration by iterating over all measurement items using gotoAnnotatedNode(),
and setting measurement properties, when applicable.

Reading SR and saving JSON is not supported yet.


fedorov:
@EmilyLindemer would you like to try out this branch? CI is failing since the conversion from DICOM to JSON is not working yet, but DICOM encoding should be fine.

If it is difficult for your to set up the build, and you would prefer to wait until docker and/or binaries are available, let me know.

PR #310, opened 2017-11-04
COMP: Fix Slicer extension packaging

jcfr:
Setting CPACK_INSTALL_CMAKE_PROJECTS variable became a requirement
following Slicer r25181. That said extensions only started to fail
packaging recently because fix introduced in r25181 was made effective
after Slicer r26488.

See https://discourse.slicer.org/t/extension-build-failures/1274/4

r25181: BUG: Extension packaging: For SuperBuild extension, ensure fixup happens last
r26488: COMP: Ensure 3rd libs and binaries are fixed up for SuperBuild extensions

http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=25181
http://viewvc.slicer.org/viewvc.cgi/Slicer4?view=revision&revision=26488

Reported-by: Andriy Fedorov <fedorov@bwh.harvard.edu>


fedorov:
Thank you JC! Sorry if this is something we missed to fix based on forum announcements.

jcfr:
> Sorry if this is something we missed to fix based on forum announcements

No worry. This had to be fixed.

At least now, a configure error will be report for other developer.

PR #317, opened 2017-12-07
Util Classes to build Measurement Report and generate JSON

michaelschwier:
These are a few classes I used to be able to conveniently put together measurement reports from custom data storage in a python batch processing script. The classes allow export to JSON which then can directly be processed by the tid1500 writer.

Currently these classes are still quite specific to my task (converting Prostate MRI Test-Retest Data/Measurements to DICOM). However I think the basic idea is nice and can be easily extended to be much more generic. The classes are mainly data structures with the same layout as the respective parts of the JSON meta file, plus some convenience methods (e.g. exporting as JSON, adding sub-items).

To make it more generic my idea would be to extend e.g. the Findings and MeasurementItem classes by proper factories that create all the typical objects by a few keywords (and also give the user a generic fallback to set all members of an object manually for cases not covered yet). In general: create convenient prepared classes so a user only has to provide the essential data/info and not worry about which special codes to populate them with.

Building this in python would also allow double use of this util library: 
1. Users can write python scripts to convert their measurement data in batch processing
2. We can also build a web-app (like the one for segmentations) on top of this lib (I think Django would be an appropriate web framework  that uses python as backend)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,7 @@ CMakeLists.txt.user*
 __pycache__/
 *.py[cod]
 *$py.class
+
+# VSCode             #
+######################
+.vscode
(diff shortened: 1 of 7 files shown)


fedorov:
@michaelschwier I actually forgot about this! This should be quite helpful for another project.

PR #334, opened 2018-02-23
WIP: add new small non-square seg test datasets

fedorov:
The dataset was generated using 3D Slicer as follows:
 * download sample CTChest dataset
 * use CropVolume module to define a small region of interest, turning off
   interpolated cropping
 * use CreateDICOMSeries to export the cropped dataset as DICOM
 * use DCMTK dcmodify to fix the PixelSpacing attributes to be within 16 characters
 * label was generated in Slicer by using Editor module Paint tool, and converted
   into DICOM SEG using itkimage2segimage


fedorov:
@hmeine can you check MeVisLab implementation whether you get the roundtrip test working for the two datasets I added in `data/segmentations/23x38x3` and `data/segmentations/24x38x3`? If you load the DICOM SEG in `seg` folder, does it line up with the one from the `nrrd` folder?

hmeine:
* I don't know what you mean with "the roundtrip test" ATM.
* When I load `data/segmentations/24x38x3`, I get one SEG file and two matching CT ROIs. According to the DICOM tags, the SEG was derived from `24x38x3/image/IMG0001.dcm`, but also shares the FoR with (one frame of) `24x38x3/multiframe/mf.dcm`. It does not matter which of the two ROI volumes I choose, both look the same.
* When I load `data/segmentations/23x38x3`, the `MultiFileVolumeListSEGOutput` module states that it appears derived from `24x38x3/image/IMG0001.dcm` as well, and does not share the FoR with any of the volumes in the `23x38x3` folder.
* Furthermore, I am seeing errors when trying to access the pixel data of `data/segmentations/23x38x3`. I will report this, but it'll take some time to be fixed, I expect, since the responsible person is on paternity leave.
* I found the nrrd files harder to read than the SEG ones, but then they had the same extent, world matrix & content. (data type was int16 instead of uint8.)

fedorov:
@hmeine thank you for testing and patience! In light of the discussion, let me re-create the sample datasets, I have some ideas how to make this more reliable.

By "round-trip test" I meant to confirm that segmentations loaded from DICOM SEG and NRRD line up. I didn't realize NRRD loading is more challenging in MeVisLab. What I plan to do next is to make a fake CT series storing the content of segmentation pixel-wise, this way the image can serve as the baseline.

The quirks of the DICOM CT datasets I shared are due to multiple factors: the multi-slice CT series was created by a very basic tool, then the result was converted using David Clunie's PixelMed toolkit, and none of the tools I use complain about the issues you pointed out.

For the new dataset, I should also put more meaningful content into SeriesDescription etc (note to myself).

fedorov:
@hmeine I updated the test objects. It should now also be a lot easier to check whether they are interpreted consistently with dcmqi/Slicer - segmentation contour in the SEG object should line up with the non-zero blob in the CT object. See visualization in Slicer below:

* 23x38x3: 23x38x3
* 24x38x3: 24x38x3

Now we just need to figure out which implementations are incorrect, since the result is not as expected in most independent implementations!

* 24x38x3: 
  * works as expected in BrainLab (see screenshot in https://github.com/fedorov/dcmqi/blob/c172992ec4a4036af5fe87bfeb7ed6fb3df381b7/data/segmentations/24x38x3/screenshots/brainlab.jpg), and in MevisLab
  * fails in Pixelmed, dcmjs
* 23x38x3:
  * fails in all independent implementations tested: BrainLab, MevisLab, dcmjs, Pixelmed

@michaelonken what are your thoughts? I definitely have not ran out of ideas how to explore this, and no non-DCMTK implementation yet said they are absolutely sure their implementation is correct, but I thought I let you know.

michaelonken:
24x38x3 looks correct when rendered through DCMTK's dcm2pnm. So I would consider the seg object correct.
23x38x3 looks broken in dcm2pnm. So probably that seg object is broken.

Either it has been fed the wrong way into dcmseg, or dcmseg has messed it up. I will investigate further within DCMTK/dcmseg.

fedorov:
@michaelonken thanks! I should have checked dcm2pnm earlier!

It looks like rendering from dcm2pnm is consistent with Brainlab rendering (below left is Brainlab, right is dcm2pnm):

image

PR #336, opened 2018-02-23
cmake: Initialize CMAKE_CXX_STANDARD to 11 if building against Slicer

jcfr:
This will allow to conveniently build dcmqi against Slicer built against
Qt5 without having to explicitly pass CMAKE_CXX_STANDARD option.


fedorov:
This made me realize that 4.8 was using `master` 😱

PR #344, opened 2018-05-15
BUG: ClinicalTrialSeriesID was never read properly from metadata

che85:
- ClinicalTrialSeriesID has always been set to `Session1` for all
  generated segmentations


fedorov:
Thanks!

PR #347, opened 2018-07-03
BUG: Fix dockcross script creation container cleanup

thewtex:
When the container is created, it does not have the name
'dockcross/manylinux-x64'. This is the name of the image as opposed to
the container. Docker gives a random, human-readable name
to a container if the `--name` flag is not provided.

Use the `--rm` flag so the container does not need to be removed.


fedorov:
@thewtex thank you so much for fixing this! 👍 I definitely don't use docker often enough! 

@jcfr FYI

Issue #352, opened 2018-08-21
Add support for encoding qualitative evaluations


fedorov:
Qualitative evaluations are supported by [TID 1411](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1411) (as well as parent and sibling templates).

Encoding of qualitative evaluations is popular in some prominent studies, such as LIDC, and as captured in the [TCIA LIDC-IDRI collection](https://wiki.cancerimagingarchive.net/display/Public/LIDC-IDRI).

PR #353, opened 2018-08-21
WIP: Add qualitative evaluations to TID 1500 converters

fedorov:
Aims to address #352.

- [x] updated schema to support metadata JSON organization
- [x] add support to tid1500writer
- [x] add support to tid1500reader

Unresolved issue that needs to be discussed is summarized below in an email sent to @dclunie today:

>Today I was working on adding support for qualitative evaluations to dcmqi, and it occurred to me that it would 1) be more consistent and 2) simplify implementation, if the qualitative assessment items (rows 16 and 17) in TID 1411 [1] were nested within a container with the heading (C0034375, UMLS, "Qualitative Evaluations"), as it is done in TID 1500 row 12 [2].
>
> As is right now, I think it is confusing since there are other CODE items in 1411 (e.g., "Finding"), and the order of items is non-significant. So I guess to parse the document, one would need to look for all CODEs that do not have concepts that otherwise can appear in the document per template structure. And then I guess there is no guarantee that someone would not report something about concept "Finding" as a Qualitative Evaluation. Needless to say, maintenance of the parser becomes a lot more complicated.
>
> I do not see a reason why it should be this way, instead of unambiguously including a container with qualitative evaluations. Unless I miss some other parsing strategy that would take care of this issue.
>
> The same issue is applicable to 1500 and 1410, both of which are also inconsistent with 1500.
> 
> [1] http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1411
> [2] http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#table_TID_1500

@jriesmeier FYI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ jobs:
       - run:
           name: build dcmqi
           command: cd docker && make dcmqi
+          no_output_timeout: 20m
       - run:
           name: test dcmqi
           command: cd docker && make dcmqi.test
--- a/apps/sr/Testing/CMakeLists.txt
+++ b/apps/sr/Testing/CMakeLists.txt
@@ -113,6 +113,15 @@ dcmqi_add_test(
     ${WRITER_MODULE_NAME}_ct-liver
   )
 
+dcmqi_add_test(
+  NAME ${READER_MODULE_NAME}_qualitative
+  MODULE_NAME ${MODULE_NAME}
+  COMMAND $<TARGET_FILE:${READER_MODULE_NAME}>
+    --inputDICOM ${MODULE_TEMP_DIR}/sr-tid1500-qualitative.dcm
+    --outputMetadata ${MODULE_TEMP_DIR}/sr-tid1500-qualitative.json
+  TEST_DEPENDS
+    ${WRITER_MODULE_NAME}_qualitative
+  )
 
 #-----------------------------------------------------------------------------
 set(MODULE_NAME tid1500)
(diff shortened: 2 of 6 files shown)


fedorov:
@dclunie response

> Even if I agreed with your argument, which I don't, we don't make these kind of structural changes to the standard once it has been finalized, since such gratuitous changes invalidate the installed base of software and archives (even if it is small).
> 
> We are not iterating towards perfection here; we have to live with what we have already done.

@jriesmeier response

>> As is right now, I think it is confusing since there are other CODE
>> items in 1411 (e.g., "Finding"), and the order of items is
>> non-significant. So I guess to parse the document, one would need to
>> look for all CODEs that do not have concepts that otherwise can appear
>> in the document per template structure. And then I guess there is no
>> guarantee that someone would not report something about concept
>> "Finding" as a Qualitative Evaluation. Needless to say, maintenance of
>> the parser becomes a lot more complicated.
>
> I agree with your observation, Andrey,
>
>> We are not iterating towards perfection here; we have to live
>> with what we have already done.
>
>This discussion reminds me to my old statement that the detection of the 
>templates used for creating an SR document is "not always sufficient".
>Also see my comment on David's blog posting from September 2013:
>
>   http://dclunie.blogspot.de/2013/09/what-template-is-that.html#comments
>
>As I wrote in my comment:
>
>   To me it still seems that the template mechanism has been defined in
>   a way that mainly facilitates the creation of SR documents, but not the
>   "consumption".
>
> David, are there any plans to enhance the existing template identification 
> mechanism in the standard in this regard?

fedorov:
Merging since the issues appear to be unrelated to the changes.

PR #360, opened 2018-10-09
ENH: add support for describing device observer context

fedorov:
This functionality was needed in particular to be able to better describe software version in https://github.com/Radiomics/pyradiomics/pull/437.

### DIFF
--- a/apps/sr/tid1500reader.cxx
+++ b/apps/sr/tid1500reader.cxx
@@ -96,13 +96,6 @@ int main(int argc, char** argv){
   doc.getInstanceNumber(temp);
   metaRoot["InstanceNumber"] = temp.c_str();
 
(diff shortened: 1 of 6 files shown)


fedorov:
@pwighton - the functionality we discussed by email is now merged, I guess the updated docker should be there soon.

What DICOM does not specify is to how the Device observer semantics maps to software, as it seems the main use case was to communicate attributes of a physical device producing the report. I am thinking to use "Device Observer Name" to encode the name of the software, and "Device Observer Model Name" for the version. I emailed @dclunie in case he has better suggestions. I think it would be nice to have the standard be more explicit about describing software, but this will not happen anytime soon, so we have to make some decisions now.

pwighton:
Thanks @fedorov, that seems reasonable.

PR #365, opened 2018-10-26
ENH: always initialize TrackingUID

fedorov:
If not provided by the user, initialize it. This can't hurt, but will come handy if measurements are done later using this segmentation.

### DIFF
--- a/libsrc/ImageSEGConverter.cpp
+++ b/libsrc/ImageSEGConverter.cpp
@@ -241,6 +241,11 @@ namespace dcmqi {
 
         if(segmentAttributes->getTrackingUniqueIdentifier().length() > 0)
           segment->setTrackingUID(segmentAttributes->getTrackingUniqueIdentifier().c_str());
+        else {
+          char dimUID[128];
+          dcmGenerateUniqueIdentifier(dimUID, QIICR_UID_ROOT);
+          segment->setTrackingUID(dimUID);
+        }
 
         CodeSequenceMacro* typeModifierCode = segmentAttributes->getSegmentedPropertyTypeModifierCodeSequence();
         if (typeModifierCode != NULL) {


pieper:
> if not initialized by the user explicitly in a longitudinal tracking situation, TrackingUID will be different even if the structure is the same

I'd say this is appropriate - if the the user doesn't explicitly affirm that the lesions are the same then the should have different UIDs.

fedorov:
Well, I think the question becomes what is the significance of something not being said. Let's say we have two time points with segmentations, and there are three situations for SEG (all of which are valid):
1. tracking UID is assigned, and is different
2. tracking UID is assigned, and it is the same
3. tracking UID is not assigned

Then let's say someone wants to do measurements over that SEG, and that someone wants to assign consistent UID for the timepoints. If option 2 or 3 above is followed, this is possible, but otherwise producer of SR is locked between two unfavorable choices: 
* violate the standard by assigning the same TrackingUID in SR (since it [says](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.4.html#table_C.8.20-4) "[Tracking UID] will have the same value as the value of the (112040, DCM, "Tracking Unique Identifier") content item in SR instances that reference this Segment in this Segmentation Instance.")
* propagate TrackingUID from SEG, and not express the longitudinal connection between the items.

The problem is SEG and SR may be created by different entities, with SEG generator not aware of consequences of not populating an optional parameter.

I am leaning towards not initializing TrackingUID in SEG by default, unknown to the user.

fedorov:
TODO:
- [ ] revise tid1500writer to check TrackingUID from SEG and propagate/issue error/warning if mismatched

jriesmeier:
@fedorov Regarding the "tid1500writer": The [setReferencedSegment()](https://support.dcmtk.org/docs/classTID1411__VolumetricROIMeasurements.html#a9fe7b153a4a9a52f30ed03b3d379de84l) method from [TID 1411 class](https://support.dcmtk.org/docs/classTID1411__VolumetricROIMeasurements.html) also sets (i.e. copies) the tracking information from the referenced segment if parameter "copyTracking" is true (which is the default). Not sure whether this information is helpful, but I just wanted to let you know.

fedorov:
Thank you Jörg, this is indeed helpful to know!

PR #368, opened 2019-01-28
ENH: update to jsoncpp 1.8.4

fedorov:
If successful, this will be a step towards replacing the custom fork of dcmqi
in MITK with the upstream


fedorov:
@nolden fyi - hopefully we will see tomorrow if this works on the Slicer dashboard

PR #369, opened 2019-01-29
COMP: Ensure dcmqi can be build with DCMQI_BUILD_APPS set to OFF

jcfr:
Co-authored-by: Andrey Fedorov <andrey.fedorov@gmail.com>


fedorov:
Thanks a lot @jcfr! Dashboard is green, as expected.

@nolden I am merging this, so you can try master tomorrow. But I also think once we confirm this solution works, I will make a release and you point to a release from MITK.

PR #374, opened 2019-01-30
COMP: Fix building of project against DCMQI when DCMQI_BUILD_APPS is OFF

jcfr:
Co-authored-by: Nolden Marco <M.Nolden@dkfz-heidelberg.de>


fedorov:
@nolden this is merged, you can test with master

PR #375, opened 2019-01-31
circleci: Ensure workflow is not triggered when latest(-tmp) branch are pushed

jcfr:


jcfr:
:tada: 

Finally, it works ... 

image


image

fedorov:
Wow, thank you JC!

But if it took JC this many iterations to fix it - it must be super-hardcore stuff....

jcfr:
> But if it took JC this many iterations to fix it [...] it must be super-hardcore stuff


To clarify, this could have been faster. the "issue" is that we had to wait a full cycle of PR, then merge, then fix ... having the temporary ability to push (or even push force) to master would have speed up this. It would have taken less than 30mins or so (indeed temporally updating the build step to create an dummy tar.gz instead of effectively building would have done the trick ...)

Issue #383, opened 2019-05-08
Test and update DCMTK version to benefit from performance enhancements


fedorov:
Following on email from @michaelonken:

> I recently looked into the performance issue with writing segmentations
> (and other objects based on functional groups).
> 
> As I thought, the main speed issue is the checking of the functional
> group structure. It's possible to speed this up (I think) by changing
> the way functional groups are organized and accessed in memory and I put
> it on my list to enhance this in the future.
> 
> For now, I did two things:
> 
> - For some objects the checking has been performed twice, including
> segmentations, i.e. updating to new code will speed up writing by
> factor ~2.
> 
> - Also, if you use the same code over and  over again to produce the
> segmentation objects, you probably do not require to check the
> correctness of the functional groups at all. In such cases it only makes
> sense to use the checking when you develop your own code and test with
> some instances.
> 
> As an impliciation I think it is useful to just optionally disable the
> checks. That's now possible too and makes writing even large
> segmentation objects very fast.
> 
> You could test this with the latest DCMTK, e.g. commit
> [9461c7851c3465ca0f14b964ed7c6d0dd4d88834](http://git.dcmtk.org/?p=dcmtk.git;a=commit;h=9461c7851c3465ca0f14b964ed7c6d0dd4d88834) (latest online right now).
> 
> The actual commit with the check option is this one:
> [591a1bc2f0ad0edec4fbb093ceb07b5fcaf127c9](http://git.dcmtk.org/?p=dcmtk.git;a=commit;h=591a1bc2f0ad0edec4fbb093ceb07b5fcaf127c9) .
> 
> You can disable checks by setting setCheckFGOnWrite(OFTrue) on your
> DcmSegmentation object.
>

fedorov:
@michaelonken can you give us some unofficial idea if a minor released is coming up sometime soon?

michaelonken:
It's planned for late summer...this year :-).

fedorov:
Resolved by #391

fedorov:
TODO: need to test with checks disabled ...

PR #391, opened 2020-01-15
ENH: update to DCMTK 3.6.5

fedorov:
re #383

### DIFF
--- a/CMakeExternals/DCMTK.cmake
+++ b/CMakeExternals/DCMTK.cmake
@@ -34,7 +34,7 @@ if(NOT DEFINED DCMTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     ${proj}_REVISION_TAG
-    "29f9de10c2e8576147f16475b156bba98d14ba7d"
+    "0f2de2313a00f9360bdf33399a2f37ee5e65c429"
     QUIET
     )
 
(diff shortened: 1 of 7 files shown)


fedorov:
@jcfr do you have any idea about the appveyor build failures?

The compiler version is the same, and I did not change anything in the cmake settings for building the dependencies (a patched version of ITK and the updated DCMTK).

fedorov:
I was hoping to update DCMTK in Slicer once I figure it out for dcmqi.

fedorov:
TODO: should experiment with disabling this flag: https://github.com/QIICR/dcmtk-dcmqi/blame/dcmqi/CMake/dcmtkPrepare.cmake#L267

fedorov:
@afshinmessiah I set `BUILD_SHARED_LIB` to OFF both for DCMTK and dcmqi, but I still see mismatched flags. I checked the DCMTK build log before the update, and that one has `/MD`, but the latest has `/MT`. Maybe we should try patching flags the same way it is done in DCMTK here: https://github.com/QIICR/dcmtk-dcmqi/blob/dcmqi/CMake/dcmtkPrepare.cmake#L330-L343. I don't know how this will affect the superbuild when dcmqi is built as part of 3D Slicer superbuild process.

fedorov:
* new release for the current master created: https://github.com/QIICR/dcmqi/releases/tag/v1.2.2
* Slicer "preview" updated to use the latest release, which is using DCMTK 3.6.3, same as Slicer: https://github.com/Slicer/ExtensionsIndex/pull/1677 (note this will need to be changed, once this PR is integrated: https://github.com/Slicer/Slicer/pull/1308)
* Slicer 4.10 release updated to use latest release: https://github.com/Slicer/ExtensionsIndex/pull/1678 (note this will remain in place, 4.10 will not be using newer versions of dcmqi due to lack of backwards compatibility)

@afshinmessiah FYI - it's a lot of moving parts, but most of it mechanical, as long as things don't break!

Issue #386, opened 2019-12-05
Update to use SNOMED CT codes in place of SNOMED RT


fedorov:

fedorov:
Notes for Afshin, who will be using this issue as an opportunity to get started with dcmqi:
* we need to update all occurrences of `(*, SRT, *)` codes to use SCT codes; hard-coded codes are used in the source code, and also in JSON files defining lists of codes
* mapping of codes between SCT and SRT is defined in this part of the standard: http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_O.html#table_O-1
* this mapping is also included in this new component of pydicom recently contributed by @hackermd: https://github.com/pydicom/pydicom/blob/master/pydicom/_snomed_dict.py

afshinmessiah:
[Here](https://gist.github.com/afshinmessiah/96f80ec9f5a2ba2cfc96161ae73a4c14) is the list of srt codes that I couldn't find any sct equivalent for.

fedorov:
@afshinmessiah as discussed, let's also see if we can replace instances like this one:

https://github.com/fedorov/dcmqi/blob/master/libsrc/ImageSEGConverter.cpp#L352

`CodeSequenceMacro("121322","DCM","Source image for image processing operation")`

to use `#define`d codes from DCMTK:

http://git.dcmtk.org/?p=dcmtk.git;a=blob;f=dcmsr/include/dcmtk/dcmsr/codes/dcm.h;h=16e3d449ba87f5d6971d695e3511a8c2062c01f2;hb=HEAD#l2397

`CODE_DCM_SourceImageForImageProcessingOperation`

This is less error prone, and more readable/maintainable.

afshinmessiah:
I've also replaced the following srt codes with their corresponding sct (srt:sct) using [this table](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_J.html). Tell me if it doesn't make sense.
srt_sct_tags = {
    'T-9200B': '41216001',
    'T-41070': '7832008',
    'T-48052': '19715009',
    'T-45526': '11279006',
    'D4-31320': '253276007',
    'T-A0193': '128320002',
    'T-47741': '86547008',
    'T-1416B': '53967007',
    'T-41040': '299716001',
    'T-48470': '195416006',
    'R-102BC': '698348000',
    'T-14183': '41313007',
    'T-49423': '714754004',
    'T-4881F': '70253006',
    'T-43126': '91760001',
[...]

PR #393, opened 2020-01-29
Replaced the SRT codes with SCT codes

afshinmessiah:
Ref issue #386. Replaced all SRT codes in any text files of the project (*.cpp, *.h, *.json, *.py, ..).


libsrc/TID1500Reader.cpp
```diff
@@ -145,8 +145,8 @@ Json::Value TID1500Reader::getMeasurements() {
               Json::Value laterality = Json::nullValue;
               if(node->getConceptName() == CODE_SCT_FindingSite)
                 laterality = getContentItem(CODE_SCT_Laterality, cursor);
-              else if(node->getConceptName() == CODE_SRT_FindingSite)
-                laterality = getContentItem(CODE_SRT_Laterality, cursor);
+              else if(node->getConceptName() == CODE_SCT_FindingSite)
```
fedorov: You should be careful replacing those codes in reader tools. In this specific example, the reader intentionally checks SRT and SCT codes, since it should be able to parse older documents that are using SRT codes. In this specific place, and in the next one, SRT codes should be preserved.

fedorov:
For completeness, can you put the script you use to do the substitution in the [util folder](https://github.com/QIICR/dcmqi/tree/master/util)?

libsrc/TID1500Reader.cpp
afshinmessiah: You mean to have both SCT and SRT or just SRT version?

fedorov: If you take a look at the functionality of the code, it should make more sense. Can you look at that specific function and see what those condition checks are doing?

fedorov:
For the reference, and to link things together, this PR is aiming to resolve issue #386.

libsrc/TID1500Reader.cpp
afshinmessiah: I checked it out. CODE_SCT_* are macros defined in DCMTK sct.h file. The previous version used to read the CODE_SRT_* from srt.h header file. There is no build error for this new version. It seems like the code is populating a json file from a DSRDocument. As you already mentioned though, it is not going to read SRT codes anymore unless we keep both srt and sct.

util/measurementReportUtils/codeSequences.py
```diff
@@ -11,13 +11,13 @@ class Finding(CodeSequence):
 
   def __init__(self, segmentedStructure):
     if segmentedStructure == "NormalROI_PZ_1":
-      super().__init__("Normal", "SRT", "G-A460")
+      super().__init__("Normal", "SCT", "17621005")
     elif segmentedStructure == "PeripheralZone":
-      super().__init__("Entire", "SRT", "R-404A4")
```
afshinmessiah: Just noticed that some SRT codes haven't been replaced. I've used [pydicom snomed dictionary](https://github.com/pydicom/pydicom/blob/master/pydicom/sr/_snomed_dict.py). looks like it is missing some codes "R-404A4".

libsrc/TID1500Reader.cpp
fedorov: Right, there is no build error. Since the code is supposed to be able to read both the new SR documents that were created using the SCT codes for those concepts, and also the new SRT codes, reader has to check both.

util/measurementReportUtils/codeSequences.py
afshinmessiah: This is pretty much the list of missing (not replaced) srt sct codes in the project:
"T-48440", "T-47741", "T-40210", "T-1416B", "T-AB000", "T-4806E", "R-102BC", "A-13510", "T-70010", "T-141A5", "T-11500", "T-49423", "T-A0193", "T-F6078", "T-D05E4", "T-4881F", "T-4882A", "T-9200B", "T-45526", "A-26A06", "R-404A4", "T-14183", "T-41040", "T-43126", "A-10044", "T-51000", "T-48470", "D4-31320", "T-48052", "A-26A08", "T-41070", "T-00317", "T-48817", "T-4942C", "T-41066", "T-1A310"

fedorov:
Can you check those against the standard here: http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_O.html#table_O-1 ?

util/measurementReportUtils/codeSequences.py
fedorov: Can you check those against the standard here: http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_O.html#table_O-1 ?

fedorov:
Also, for commits like below:

image

it is better to [squash them into a single commit](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/), to keep history more concise.

util/measurementReportUtils/codeSequences.py
afshinmessiah: Checked the page, it doesn't have either. There is one other [page](http://dicom.nema.org/medical/dicom/2016b/output/chtml/part16/sect_CID_3010.html) though, that has some of them.

fedorov: Please work against [current version of the standard](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_3010.html). The one you referenced is from 2016.

afshinmessiah: Well, the reason that Google found the older versions was that they are deprecated values. I could find majority of missing labels [here](http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_J.html).

fedorov:
@afshinmessiah can you clean up the commit history a bit? I think the three commits starting with "Added the PYTHON script for replace srt wiht" can be merged into a single one, and also there are typos in the commit messages. You can read how to do this here: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History (just be careful!).

afshinmessiah:
Sure @fedorov, just did. Can't figure why appveyor fails.

fedorov:
There is a merge conflict for this PR - this might be the reason why appveyor complains.

Issue #395, opened 2020-02-09
Build fails on mac due to iconv undefined symbols


pieper:
While testing https://github.com/QIICR/dcmqi/issues/394, I also found that the current master isn't building on mac (10.14.6 mojave).  Same machine builds Slicer fine, so there's something off in the dcmqi superbuild configuration.

```
[  2%] Linking CXX executable ../../bin/ofstd_tests
Undefined symbols for architecture x86_64:
  "_iconv", referenced from:
      OFCharacterEncoding::Implementation::convert(OFString&, char const*, unsigned long) in libofstd.a(ofchrenc.cc.o)
  "_iconv_close", referenced from:
      OFCharacterEncoding::Implementation::~Implementation() in libofstd.a(ofchrenc.cc.o)
  "_iconv_open", referenced from:
      OFCharacterEncoding::Implementation::create(OFString const&, OFString const&, OFCondition&) in libofstd.a(ofchrenc.cc.o)
  "_iconvctl", referenced from:
      OFCharacterEncoding::Implementation::getConversionFlags() const in libofstd.a(ofchrenc.cc.o)
      OFCharacterEncoding::Implementation::setConversionFlags(unsigned int) in libofstd.a(ofchrenc.cc.o)
  "_locale_charset", referenced from:
      OFCharacterEncoding::Implementation::getLocaleEncoding() in libofstd.a(ofchrenc.cc.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[5]: *** [bin/ofstd_tests] Error 1
make[4]: *** [ofstd/tests/CMakeFiles/ofstd_tests.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [DCMTK-prefix/src/DCMTK-stamp/DCMTK-build] Error 2
make[1]: *** [CMakeFiles/DCMTK.dir/all] Error 2
make: *** [all] Error 2
```

fedorov:
I remember seeing something like this before. I think there is some conflict if you have a locally installed iconv - it's picked up by cmake erroneously. I managed to find that discussion - can you look if that helps? https://github.com/Slicer/SlicerBuildEnvironment/issues/10

pieper:
> I managed to find that discussion - can you look if that helps? Slicer/SlicerBuildEnvironment#10

I skimmed through but didn't see a solution.  Was there a particular thing you thought we should try?

fedorov:
[...]

PR #397, opened 2020-02-12
COMP: turn of SNDFILE to fix include path

pieper:
Fixex #395

As described in the issue, on some (many?)
systems dcmtk header files could end up
in the same path (e.g. /usr/local/include)
with other packages, in this case
libsndfile's header.


fedorov:
Thank you @pieper!

Should it also be disabled for Slicer?

pieper:
> Should it also be disabled for Slicer?

Yes, I will check it in when I finish a few calls.

PR #401, opened 2020-05-19
ENH: add cardinality constraint to the schemas

### LINKED ISSUE #400: Update seg-schema.json with cardinalities
In https://github.com/razorx89/pydicom-seg/issues/14 the author relied on the schema validation, which succeeded, but in the end produced the wrong result. According to the examples in this repository, e.g. [example_multiple_segments_reordered.json](https://github.com/QIICR/dcmqi/blob/master/doc/examples/seg-example_multiple_segments_reordered.json), a segment object is encapsulated into an array. Thus the outermost array, the `segmentAttributes` key, contains arrays, each having a single object. The schema, however, does not explicitly state this structure and thus a user could write the json with having the outermost array containing only a single array, which contains multiple objects.

https://github.com/QIICR/dcmqi/blob/9c0c7bf75bb89c92f5d4aa120c5f6d13404f0fc8/doc/schemas/seg-schema.json#L94-L99

Since the schema is written for jsonschema v4, it is easy to define the required cardinalities: 
https://json-schema.org/understanding-json-schema/reference/array.html#length

```
 "segmentAttributes": { 
   "type": "array", 
   "minItems": 1,
   "items": { 
     "type": "array", 
     "minItems": 1,
     "maxItems": 1,
     "items": { 
       "type": "object", 
```

This change would require the outer array to have `1..*` items and the innermost to have exactly `1` item.

fedorov:
resolves #400

### DIFF
--- a/apps/seg/itkimage2segimage.cxx
+++ b/apps/seg/itkimage2segimage.cxx
@@ -61,8 +61,9 @@ int main(int argc, char *argv[])
   metainfoisstream >> metaRoot;
 
   if(metaRoot.isMember("segmentAttributes")){
-    if(metaRoot["segmentAttrubutes"].size() != segImageFiles.size()){
+    if(metaRoot["segmentAttributes"].size() != segImageFiles.size()){
       cerr << "Error: number of items in the \"segmentAttributes\" metadata array should match the number of input segmentation files!" << endl;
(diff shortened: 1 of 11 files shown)


fedorov:
TODO:
- [ ] add readme

fedorov:
@razorx89 apparently, still need to debug this, but if you let me know about the example part, I will just merge when I am done.

PR #407, opened 2020-10-23
BUG: Fix loading of single-frame segmentation

lassoan:
When a single frame was loaded, sliceExtent was left uninitialized (instead of set to 0.0), which prevented computation of number of frames.

Fixes loading of file mentioned in https://discourse.slicer.org/t/loading-dicom-segmentation/14139/4


fedorov:
Thank you for the fix! I think you meant "prevented **correct** computation of the number of frames".

Issue #348, opened 2018-08-07
E: PatientName (0010,0010) violates VR definition in PatientModule but actual value is correct 


mmromero:
Dear developers, 

I am trying to create the structure report of a white matter hyper-intensity lesion segmentation with your docker container but no matter the Patient's Name I always get the same error (see below). The actual value is "N2D_PATIENT", but also tried with "xxx xxx" with the same result. 

I can share the data with you, just let me know if you need it and I upload it. 

Thank you very much.
Miguel.

```
docker run -v /Users/miguel/Workspace/RAPIDS/SR/dcmqi_test:/tmp/dcmqi qiicr/dcmqi itkimage2segimage --inputDICOMDirectory /tmp/dcmqi/input_data --inputMetadata /tmp/dcmqi/wmh_segmentation.json --inputImageList /tmp/dcmqi/output_nifti/18991230_000000nifti2dicom044s001a001.nii.gz --outputDICOM /tmp/dcmqi/wmh_segmentation.dcm
dcmqi repository URL: git@github.com:QIICR/dcmqi.git revision: 72869c3 tag: latest-2-g72869c3
Searching recursively /tmp/dcmqi/input_data for DICOM files
Input image size: [240, 240, 48]
W: PatientName (0010,0010) violates VR definition in PatientModule
W: PatientSex (0010,0040) violates VR definition in PatientModule
W: AccessionNumber (0008,0050) violates VR definition in GeneralStudyModule
W: InstitutionName (0008,0080) violates VR definition in GeneralEquipmentModule
W: PatientAge (0010,1010) violates VR definition in PatientStudyModule
W: PositionReferenceIndicator (0020,1040) absent in FrameOfReferenceModule (type 2)
Directions: 0.997303 -0.0502721 -0.0534678
-0.0229997 -0.905936 0.422789
0.0696929 0.42042 0.904649

Processing input label Image (0x1889550)
  RTTI typeinfo:   itk::Image<short, 3u>
  Reference Count: 3
  Modified Time: 184
  Debug: Off
  Object Name: 
  Observers: 
    none
  Source: (none)
  Source output name: (none)
  Release Data: Off
  Data Released: False
  Global Release Data: Off
  PipelineMTime: 48
  UpdateMTime: 183
  RealTimeStamp: 0 seconds 
  LargestPossibleRegion: 
    Dimension: 3
    Index: [0, 0, 0]
    Size: [240, 240, 48]
  BufferedRegion: 
    Dimension: 3
[...]

PR #409, opened 2020-11-06
ENH: Added detailed logging option to itkimage2segimage and segimage2itkimage

lassoan:
I had really hard time figuring out why segmentation export fails, and it turned out that the in the input DICOM images ProcedureCodeSequence was defined but did not contain any items, and because of this DCMTK refused to create the segmentation object. It took several hours of debugging to investigate this issue.

Along the way, I realized that DCMTK provides pretty detailed logging of all relevant details, but this logging is not enabled. To prevent such tedious investigations in the future, I added a --verbose command-line switch to segmentation converters, which enable printing of detailed information. It prints debug-level messages, which can be quite detailed, so by default the switch is disabled.

### DIFF
--- a/apps/seg/itkimage2segimage.cxx
+++ b/apps/seg/itkimage2segimage.cxx
@@ -6,6 +6,9 @@
 #include "dcmqi/ImageSEGConverter.h"
 #include "dcmqi/internal/VersionConfigure.h"
 
+// DCMTK includes
+#include <dcmtk/oflog/configrt.h>
+
 typedef dcmqi::Helper helper;
 
 int main(int argc, char *argv[])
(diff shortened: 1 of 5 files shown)


pieper (approved):
LGTM 👍

fedorov:
@lassoan is this a Windows issue? For example, looking at this issue https://github.com/QIICR/dcmqi/issues/348, the warnings/errors are already being shown on the console. Or is this PR adding even further error reporting? I am not sure I have a problematic dataset handy to test this. 

But I do not object to the PR - thank you for this contribution!

lassoan:
No DCMTK logs appear on Windows by default. DCMTK uses log4cplus, which can may initialize logging settings from some system defaults, which may explain why some logs appeared on that configuration. The default configurator enables debug-level logging, so it provides more details anyway.

You can reproduce the problem if you want to export a segmentation for the CT volumein TCIA's TCGA-CV-A6JN data set. I also copied the input files here: https://1drv.ms/u/s!Arm_AFxB9yqHxKRuA_mba8epTDT0gQ?e=DzHlOs

PR #415, opened 2021-01-27
BUG: increase precision

### LINKED ISSUE #414: DCM Seg: Rounding of orientation matrix when storing seg
It seems that DCMQI rounds values at the 5th decimal position, if storing DCM Segs.

We had this issues now with some data sets (see https://phabricator.mitk.org/T28127), where the rounding became obvious.
E.g.
Stored as nrrd:
`space directions: (0.68359375,0,0) (0,0.68359375,0) (0,0,1)`
Stored as DCM Seg (via DCMQI):
`space directions: (0.68359380000000003,0,0) (0,0.68359380000000003,0) (0,0,1)`

Before we just lower the precision of consistency checks on our side, we thought that it might make sense to check that DCMQI provides the highest possible fidelity when storing the data. That's why we wanted to point out that finding here and discuss it.

What could/is the reason for that rounding? Any idea?

fedorov:
resolves #414

### DIFF
--- a/libsrc/Helper.cpp
+++ b/libsrc/Helper.cpp
@@ -122,10 +122,11 @@ namespace dcmqi {
     // somehow this was resulting in more than 16 characters in local tests
     ostringstream sstream;
     sstream.imbue(std::locale::classic());
-    sstream << setprecision(15) << f;
+    sstream << setprecision(14) << f;
     formatted_f = sstream.str();
     cout << "Formatted float: " << formatted_f << endl;
     cout << "Size: " << formatted_f.size() << endl;
+
     /*
     ostringstream sstream;
     sstream << fixed << setprecision(15) << f;
--- a/libsrc/Helper.cpp
+++ b/libsrc/Helper.cpp
@@ -103,9 +103,32 @@ namespace dcmqi {
 
 
   string Helper::floatToStrScientific(float f) {
+    /*
+    Alternatife with scientific to get to 16:
+     - mantissa sign (1)
+     - mantissa leading number (1)
+     - dot (1)
+     - mantissa after the dot (8)
+     - E (1)
+     - exponent sign (1)
+     - exponent (2 OR 3 (Win))
+
     ostringstream sstream;
-    sstream << scientific << f;
-    return sstream.str();
+    sstream << setprecision(8) << scientific << f;
+    cout << setprecision(8) << scientific << f << endl;
+    */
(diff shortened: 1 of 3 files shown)


libsrc/Helper.cpp
```diff
@@ -103,9 +103,32 @@ namespace dcmqi {
...
+     - exponent (2 OR 3 (Win))
+
+    ostringstream sstream;
+    sstream << setprecision(8) << scientific << f;
+    cout << setprecision(8) << scientific << f << endl;
+    */
     ostringstream sstream;
-    sstream << scientific << f;
-    return sstream.str();
+    sstream << fixed << setprecision(15) << f;
```
rfloca, line 121: sstream.imbue(std::locale::classic());
before using the sstream.

libsrc/Helper.cpp
```diff
@@ -103,9 +103,32 @@ namespace dcmqi {
...
+    string formatted_f = sstream.str();
+    size_t dot_position = formatted_f.find('.');
```
rfloca, line 123: You can move thin into the if statement as it is only used in that scope.

libsrc/Helper.cpp
```diff
@@ -103,9 +103,32 @@ namespace dcmqi {
...
+    if(formatted_f.length()>16){
+      if(dot_position == string::npos || dot_position > 15){
+        cerr << "ERROR: Failed to convert " << f << " to DS VR!" << endl;
+      } else {
+        formatted_f = formatted_f.substr(0,16);
+      }
+    }
```
rfloca, line 130: Should not be needed, if everything works according to standard. But it doesn't hurt, if you have no very high performance requirements.

rfloca (changes requested):

kislinsk:
Here's my shot: http://cpp.sh/72ipr

To squeeze every little possible bit of precision out of it, you could print both scientific and fixed notation to strings and compare their lengths. If the scientific notation fits, use it. Fixed notation otherwise. The numeric_limits::max_digits10 ensures that both notations represent the maximum possible precision.

kislinsk:
Nevermind, I just noticed that DCMQI only used float so it makes things easier in the context of 16 characters. The standard library provides useful constants for serializing floating point numbers with maximum precision and for float (with all the extras) it always fits even into less than 16 characters. I will create a pull request for review.

fedorov:
replaced by #416

Issue #419, opened 2021-02-10
Crash converting fractional segmentation frame to ITK


lorteddie:
While handling fractional frames in ImageSEGConverter::dcmSegmentation2itkimage the original frame is copied. This actually is a shallow copy so both frames point to the same pixel buffer. Deleting the shallow copy deletes the buffer. Afterwards the original frame is deleted, due to the changes in #417, causing a double delete which most likely results in a crash.

The shallow frame copy is actually unnecessary, a deletion of the 'working' frame is only necessary if it is an unpacked binary frame.

fedorov:
Thank you for testing this and contributing a patch. I don't think dcmqi was used much with fractional segmentations. Would also be great to have a regression test for a fractional seg!

lorteddie:
Not sure if I find some time to go there any time soon. But I can give some feedback:
We are currently using dcmSegmentation2itkimage to convert SEG objects, binary and fractional. In general this works fine, couldn't find issues converting or parsing issues. There should definitely be an overload taking a DcmSegmentation object though. Problem here is the current interface requires us to parse the DcmSegmentation object twice.
We decided against using itkimage2dcmSegmentation because it is too limiting with its interface and the hard coded values behind it. Had to write that ourselves. And some of those limitations even proceed to DCMTK, like for example the CodeSequenceMacro which is missing half the attributes. Also the SEG standard is ridiculously limiting, for example the stupid amount of single item sequences which are a pain ...

fedorov:
Thanks for the feedback and your contribution!

> We decided against using itkimage2dcmSegmentation because it is too limiting with its interface and the hard coded values behind it. Had to write that ourselves. And some of those limitations even proceed to DCMTK, like for example the CodeSequenceMacro which is missing half the attributes. 

Can you elaborate a bit on this? You mean various type 1C and 3 in the below that you would want to populate?

![image](https://user-images.githubusercontent.com/313942/107580164-28f77480-6bc4-11eb-8e95-5c974db0ed7e.png)

lorteddie:
[...]

PR #420, opened 2021-02-10
#419 do not delete a shallow copied frame, only delete if necessary

lorteddie:


fedorov:
Looks good to me, thanks!

PR #423, opened 2021-02-12
dcmqi target install step

lorteddie:


lorteddie:
kind of trying things out here, it works for my setup but i don't ever build superbuild

fedorov:
@rfloca @kislinsk do you mind checking if this will work with MITK superbuild?

rfloca:
Yes sure.

fedorov:
@lorteddie can you let me know if/when this PR is ready for testing? Also, would it make sense to squash some commits, or you think each one deserves to be preserved?

lorteddie:
yes I redid the commits, was definitely too much trial and error commits
I tested that dcmqi target etc is not installed if build via superbuild because cmake will complain that include directories are pointing into the build directory because of ITK and DCMTK. And I tested that it is actually installed if build separately which makes the cmake integration into other projects a lot easier.
If it works for MITK I think we are good here.

fedorov:
Great, thank you! I am going to merge, and see how Slicer dashboard reacts to this. We can revert back if there are issues with Slicer or MITK identified.

rfloca:
Hi, checked it with our current develop branch. It builds and the quick smoke test with our installer also seems to work. If we should still find problems we let you know.

Issue #429, opened 2021-02-26
Combining dcmqi with Conan ITK and DCMTK packages


akalali:
I'm currently in the process of migrating our code base using [Conan](https://github.com/conan-io/conan) for third-party dependency management.
As a test project I use the already existing Conan packages of [ITK](https://github.com/conan-io/conan-center-index/tree/master/recipes/itk) and [DCMTK](https://github.com/conan-io/conan-center-index/tree/master/recipes/dcmtk): I download them from [ConanCenter](https://conan.io/center/) using some Conan-calls inside my CMakeLists.txt and use them to point `ITK_DIR` and `DCMTK_DIR` when building dcmqi with CMake's ExternalProject.

However, I'm struggling with different things so I opened an issue [here](https://github.com/conan-io/conan/issues/7636). One main point is that the `UseITK.cmake`-file is not found / used by the Conan ITK package so dcmqi fails [here](https://github.com/QIICR/dcmqi/blob/master/CMakeLists.txt#L274). I wonder what the use of this file is and why it is necessary.

I would love to bring dcmqi to Conan, but in order to create a Conan recipe file for dcmqi I need to understand the process how dcmqi is built and for that I'm going one step back and try to built dcmqi as an external project, while using Conan packages for its dependencies.

fedorov:
@jcfr any chance you can comment on this?

fedorov:
> One main point is that the UseITK.cmake-file is not found / used by the Conan ITK package so dcmqi fails here. I wonder what the use of this file is and why it is necessary.

@akalali the purpose of this file appears to be documented here:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/UseITK.cmake#L13-L16

akalali:
> 
> 
> > One main point is that the UseITK.cmake-file is not found / used by the Conan ITK package so dcmqi fails here. I wonder what the use of this file is and why it is necessary.
> 
> @akalali the purpose of this file appears to be documented here:
> 
> https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/UseITK.cmake#L13-L16

I'm wondering if there is another scenario where the source-directory and the `UseITK.cmake`-file are not existent and only the build-folder (bin, include,lib) can be used for including ITK as a third-party dependency? What will happen in that case with the `CMakeLists.txt`-file of dcmqi?

fedorov:
[...]

PR #432, opened 2021-03-12
Dcmqi install step improvements

lorteddie:


fedorov:
@akalali FYI - in case you want to check this in the context of #429

fedorov:
@lorteddie are you still working on this, or it is ready to merge?

lorteddie:
ready from my side, unfortunately i didn't test the last cmake change after removing the build directory, this change fixes paths in the cmake files pointing into the build directory
i also removed the zlib include since i don't understand why it is there (and it fails with dcmtk built with zlib). and since the dcmtk cmake config file still uses library and include paths instead of targets, their dependencies like zlib aren't covered by DCMTK_INCLUDE_DIRS

fedorov:
I will merge, but will see how this plays with the Slicer superbuild.

Issue #446, opened 2022-03-04
Test transition to C++ 14


fedorov:
Currently, dcmqi fails in Slicer:

https://slicer.cdash.org/build/2599227/configure

<img width="800" alt="image" src="https://user-images.githubusercontent.com/313942/156803141-667434bd-a807-4bad-9179-3bf807c70428.png">

fedorov:
Need to implement the fix as in https://github.com/acil-bwh/ChestImagingPlatform/pull/42

fedorov:
This was addressed by https://github.com/QIICR/dcmqi/pull/458

PR #448, opened 2022-03-04
COMP: will it work with C++ 14?

fedorov:


fedorov:
hopefully, resolves #446

PR #458, opened 2022-07-05
COMP: Fix Slicer extension build by not forcing CMAKE_CXX_STANDARD value

jcfr:
This commit fixes the following error introduced by
Slicer/Slicer@dedb6c04d5 (COMP: Require at least C++17 (PR-6237))

```
  -- Checking if building as a Slicer extension - yes (DCMQI_BUILD_SLICER_EXTENSION is ON)
  CMake Error at /work/Preview/Slicer-0-build/Slicer-build/SlicerConfig.cmake:969 (message):
    Variable CMAKE_CXX_STANDARD defined prior calling 'find_package(Slicer)'
    does NOT match value used to configure Slicer.  It probably means that a
    different CMAKE_CXX_STANDARD has been used to configure this project and
    Slicer.

    CMAKE_CXX_STANDARD=14

    Slicer_CMAKE_CXX_STANDARD=17
  Call Stack (most recent call first):
    /work/Preview/Slicer-0-build/Slicer-build/SlicerConfig.cmake:983 (slicer_config_set_compiler_ep)
    CMakeLists.txt:233 (find_package)
```

cc: @sjh26 @jamesobutler


CMakeLists.txt
```diff
@@ -6,7 +6,9 @@ cmake_minimum_required(VERSION 3.5.0)
 set(_msg "Setting C++ standard")
 message(STATUS "${_msg}")
 if(DEFINED Slicer_DIR)
-  set(CMAKE_CXX_STANDARD 14)
+  if(NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+  endif()
 endif()
 if(NOT DEFINED CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 98)
```
jamesobutler, line 14: Not a blocker to merging this change, but if a `CMAKE_CXX_STANDARD` is not set, dcmqi wants to still continue setting the `CMAKE_CXX_STANDARD` to `98` for a non-Slicer dir and to `14` for a Slicer dir? Does dcmqi want to standardize what C++ standard it should conform to?

fedorov: Would you recommend we switch to `14` in all places as default?

jamesobutler: The current Slicer stable will build as `14` while Slicer preview will build as `17`. If not a `Slicer_DIR`, then the minimum of Slicer stable and preview would seem to make sense. Therefore setting it to `14` when not a `Slicer_DIR` would make sense.

In terms of general relevancy, latest versions of many projects are now switching to `17`.

fedorov: Thank you @jamesobutler!

@jcfr let me know if you prefer to merge this PR and handle this separately, or want to update in place. Either is fine with me.

jcfr:
Thanks @fedorov :pray:

fedorov:
Thank YOU @jcfr!

PR #459, opened 2022-07-19
support for new dcmtk targets

lorteddie:
DCMTK 3.6.7 introduced targets like DCMTK::dcmdata which should be used instead of the old style listing all modules and include directory

### DIFF
--- a/libsrc/CMakeLists.txt
+++ b/libsrc/CMakeLists.txt
@@ -64,8 +64,17 @@ endif()
 
 set_property(GLOBAL APPEND PROPERTY ${CMAKE_PROJECT_NAME}_TARGETS ${lib_name})
 
+set(_dcmtk_libs)
+set(_dcmtk_includes)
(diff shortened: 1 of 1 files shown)


fedorov:
> it should work with with both versions

Ok, if it should work, that's fine. I was confused since you said "3.6.7 introduced targets like DCMTK::dcmdata which should be used instead of the old style", and I understood it that this is a feature specific to 3.6.7.

lorteddie:
oh sorry, i meant they should be preferred over the old style. the changed code tests if the new style dcmtk targets are available by just testing if dcmtk::dcmtk is available as target, dcmtk:dcmtk is the new 'all modules' target pretty much equivalent to the old 'DCMTK_LIBRARIES' variable which listed all modules as a list. if the new style is there use it, if not fallback to old style.

fedorov:
I see it, sorry I missed this initially. Thank you for the contribution!

PR #466, opened 2023-01-26
BUG: Ensure binaries bundled into dcmqi docker image can run

jcfr:
This commit is a follow-up of eca78a5c5 (replace deprecated dockcross image) updating the build environment from `dockcross/manylinux-x64` to `dockcross/manylinux2014-x64`.

Since `dockcross/manylinux2014-x64` is based on Centos 7, this commit also updates the base distribution associated with the dcmqi docker image to be consistent.

See https://github.com/pypa/manylinux#manylinux2014-centos-7-based


jcfr:
After CircleCI completes, I will push an additional commit adding regression tests to ensure
executable bundled in the docker image can successfully run.

PR #476, opened 2023-09-12
Fix Windows Release publishing.

michaelonken:
Use powershell multi-line command separators (backticks) for long command line call.

### DIFF
--- a/.github/workflows/cmake-win.yml
+++ b/.github/workflows/cmake-win.yml
@@ -67,10 +67,10 @@ jobs:
       - name: Publish package
         run: |
           pip install -U "scikit-ci-addons>=0.22.0"
-          ci_addons publish_github_release qiicr/dcmqi 
-           --exit-success-if-missing-token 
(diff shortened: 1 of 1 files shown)


fedorov:
Oh wow! I did have those backticks, but I removed them, since I thought they are an artifact of Appveyor syntax. Thank you for catching this! Not sure why it is failing now.... We also should look (at some point) how to disable package generation and upload when the workflow is triggered on PR.

michaelonken:
I suspect that windows build above fails since the [secret.GA_TOKEN](https://github.com/michaelonken/dcmqi/blob/64abb2246e67ace5b7a17150fd94ec0436854492/.github/workflows/cmake-win.yml#L76C1-L76C42) is probably not accessible from my pull request (see also [same error on Linux build for this PR](https://github.com/michaelonken/dcmqi/blob/64abb2246e67ace5b7a17150fd94ec0436854492/.github/workflows/cmake-win.yml#L76C1-L76C42)).

I tried on my own repo with my own token and it seems to be fine in general.

fedorov:
I am merging this one, since overall PRs do not need to be generating packages, so we can ignore those errors for now I guess.

Issue #482, opened 2023-11-17
Fix tagged release creation and package upload


fedorov:
Those have been broken since we migrated to GitHub actions :-( Summary below is what I communicated to Vamsi by Discord, but it is better to continue the discussion here for transparency and coordination, so I replicate what I said below.

Some time ago I migrated dcmqi to GA from  Appveyor/TravisCI/CircleCI. That was my one and only major foray into GA. I think I got most of the things figured out, with one exception. We used to have the process where tagged git commits that follow the versioning convention would be uploaded to GitHub and a new release would be created, but that got broken, and I feel it would take me too much time to figure out how to fix it.

The original CI was set up by Jean-Christophe @jcfr from Kitware (whom you will hopefully meet at the project week!), and he used a dedicated python package to handle this, see https://github.com/QIICR/dcmqi/blob/master/.github/workflows/cmake-linux.yml#L60-L65. I think the motivation for that package was to deal with the differences across CI platforms.

Packaging and upload of the "latest" release works fine, but not the tagged releases, which I believe would be handled by this clause in the command: `--release-packages "build/dcmqi-build/dcmqi-*-linux.zip"`

This is the commit that I tagged, which would be expected to generate and upload a release package: https://github.com/QIICR/dcmqi/commit/697ff258b6cc76fcf04d7e1dd35dc67e08f8b857
but it only uploaded latest: https://github.com/QIICR/dcmqi/actions/runs/6671898376/job/18134744092

It may be as simple as the package not assigned the expected file name. As you can see, the package file is named dcmqi-1.2.6-linux-20231027-697ff25.tar.gz, and the release pattern is dcmqi-*-linux.zip.  Which as I just now realize can't be right - it should have tar.gz extension. But it also failed on all other platforms to upload the tagged release package.

The reason I would really like to fix it is because yesterday we merged the dcmqi improvement that speeds up DICOM SEG read by orders of magnitude, and I would want to have a release to make this capability easily accessible ("latest" packages are not persistent, so it makes no sense pointing users to include them into notebooks etc).

fedorov:
I should also correct myself - it was not just me who did the migration to GitHub actions - Michael Onken @michaelonken contributed a lot to this as well, helping fixing bugs and addressing many of the related issues as well! 🙏

vkt1414:
Happy to work on this! I'll first try to troubleshoot with scikit itself. And if you are okay with using github api, you may consider the following.

[...]

PR #483, opened 2023-11-19
Add fetch tags to exisitng workflows

vkt1414:
To address #482, I found that the reason for scikit unable to make releases with tagged commits (w/ versioned tags). The git checkout action does not fetch tags unless we explicitly mention it. Some discussion can be seen here..https://github.com/actions/checkout/issues/206
I came to know about this issue when I simply wanted to see the tags while running the GitHub actions workflow with 'git describe' and the logs said no tags found. I then searched if this is a common issue and found the above discussion.
image

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -37,21 +37,21 @@ jobs:
       run: |
         cd docker && make dcmqi.test
 
-    - name: "Publish docker image"
-      # Only run if the event is not a pull request and the repository owner is QIICR.
-      # The latter is to prevent forks from publishing packages even if the owner's token
-      # would have sufficient privileges.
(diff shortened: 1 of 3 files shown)


vkt1414:
update: I modified some minor changes to expected release file names and everything should now work as expected.

fedorov:
Very nice! How did you figure this out?! Thank you, this is very helpful! I will make a release tag a bit later to check if this indeed works.

I am going to squash the commits to keep history more clean. Next time you can do `git rebase -i ` (see [here](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)) and force-push.

vkt1414:
Thanks, @fedorov for the tip. I absolutely need to learn more git tricks, particularly git rebase I guess as on more than one occasion I wanted to simplify the commit history.

The discussion here about tags helped solve (hopefully) this issue..https://github.com/actions/checkout/issues/206

PR #485, opened 2023-12-08
Use label id as segment numbers

michaelonken:
This pull request contains two parts:

**New functionality:**
- In ITK to DICOM conversion, use Label IDs from input files as Segment  Numbers if possible. From the related API documentation:
```
     * @param sortByLabelID A boolean indicating whether to sort the segments by label ID.
     *        In this case, the label IDs are used as DICOM segment numbers. This only works
     *        if the label IDs start at 1 and are numbered monotonically without gaps. The processing
     *        order of label IDs is not relevant, i.e. they can occur in any order in the input.
     *        If n labels are not assigned uniquely to label IDs 1..n in the input, the
     *        conversion will fail.
     *        If this is set to false (default), the segment numbers are assigned in the order of the
     *        labels that are being converted, i.e. the first label will receive the Segment
     *        Number 1, the second label will receive the Segment Number 2, etc.
```
Also, a new test group is added that uses some (more than before) sophisticated segment setup with overlapping segments. A first test produces a DICOM file from three NRRD inputs, which are merged together in one DICOM file. A second test splits them again into 3 NRRD files (where 2 NRRD files will contain more than one segment) and compares them to the original inputs from test 1. A third test compares metadata input used in test 1 to metadata output produced by test 2.
     
**Various small enhancements in ITK to DICOM conversion context:**
- Replaced some deprecated JSON writer code
- Enhanced documentation
- Made logging output a bit more consistent
- Save some memory (don't copy segmentation dataset on return)


libsrc/Itk2DicomConverter.cpp
```diff
@@ -285,6 +291,7 @@ namespace dcmqi {
 
         Uint16 segmentNumber;
         CHECK_COND(segdoc->addSegment(segment, segmentNumber /* returns logical segment number */));
+        segNum2Label.insert(make_pair(segmentNumber, label));
```
fedorov, line 294: Here, different values of `segmentNumber` can map to the same value of `label`, which may not be unique across input segmentation files. Is this ok?

fedorov: Yes, looking further, this will definitely be a problem, since this map is later used for renumbering segments!

libsrc/Itk2DicomConverter.cpp
```diff
@@ -483,10 +494,117 @@ namespace dcmqi {
         segmentsOverlap = "NO";
       else
         segmentsOverlap = "UNDEFINED";
-      CHECK_COND(segdocDataset.putAndInsertString(DCM_SegmentsOverlap, segmentsOverlap.c_str()));
+      CHECK_COND(segdocDataset->putAndInsertString(DCM_SegmentsOverlap, segmentsOverlap.c_str()));
+    }
+
+    if (sortByLabelID)
+    {
+      sortByLabel(segdocDataset.get(), segNum2Label);
```
fedorov, line 502: I do not see how this could possibly work. I do not see any checks for non-consecutive segment numbering, so if there are gaps in label numbers, the result will violate the conventions for having consecutive values for `SegmentNumber`. Did I miss something?

fedorov (changes requested):
I spent quite some time with this PR, and I have a lot of questions. I am not sure it will work.

Also, note that `MultiFrameDimensionModule` utilizes `ReferencedSegmentNumber` in the per-frame FGs, see: https://github.com/QIICR/dcmqi/blob/master/libsrc/Itk2DicomConverter.cpp#L58-L62.

I would think that all of the frames would need to be reordered in the general case if the segments are renumbered.

Let me know if it is easier to have a screen share to go over this!

michaelonken:
Thank you for the thorough review @fedorov!
I will check soon and provide feedback.

-- 
Sent from mobile

14.12.2023 22:35:48 Andrey Fedorov ***@***.***>:

libsrc/Itk2DicomConverter.cpp
fedorov: Maybe an easy thing to do is to check if label value is present in the map prior to inserting it, and if it is indeed present, reset `sortByLabelID` to `false`, since sorting becomes impossible?

apps/seg/itkimage2segimage.xml
```diff
@@ -64,6 +64,15 @@
       <description>Skip empty slices while encoding segmentation image. By default, empty slices will not be encoded, resulting in a smaller output file size.</description>
     </boolean>
 
+    <boolean>
```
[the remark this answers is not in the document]
michaelonken: I used this approach for two reasons:

1. Don't break backward compatibility
2. Make sure that conversion succeeds also in cases where label IDs cannot be sorted (since then the call will fail). I assume the user does often not know beforehand, whether the labelmaps will allow sorting. So the idea is that you only turn on this functionality if the user knows the prerequisites for this are fulfilled.

If 1. does not matter, one could also enable it by default and fall back to the old behavior if re-sorting does not work. I am not sure whether this can be implemented well, since the "rollback" after half of the re-resorting is probably difficult. And finding this out before actually starting the sorting work costs extra time.

libsrc/Itk2DicomConverter.cpp
```diff
@@ -367,16 +372,16 @@ namespace dcmqi {
               CHECK_COND(fgder->addDerivationImageItem(CodeSequenceMacro(code_seg.CodeValue,code_seg.CodingSchemeDesignator,
 				  code_seg.CodeMeaning),"",derimgItem));
 
-              //cout << "Total of " << siVector.size() << " source image items will be added" << endl;
+              // cout << "Total of " << siVector.size() << " source image items will be added" << endl;
```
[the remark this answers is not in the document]
michaelonken: Yes, sure. There are several places IIRC where I was not sure what to do with old "debug-like" commands. If you say I can do what I find useful, I will over all log messages and try to bring them into good shape.

libsrc/Itk2DicomConverter.cpp
michaelonken: I assumed that labels are unique in the input. If this assumption does not hold, then indeed the sorting does not make sense (since the idea to use the original labels as Segment Numbers cannot work then if two labels use the same number)

libsrc/Itk2DicomConverter.cpp
michaelonken: I think you are right, this is not addressed.  Probably it can easily by done by either checking the labels in in segNum2Label, or iterating over the segments in Segment Sequence once mapping has been performed.

michaelonken:
> Also, note that `MultiFrameDimensionModule` utilizes `ReferencedSegmentNumber` in the per-frame FGs, see: https://github.com/QIICR/dcmqi/blob/master/libsrc/Itk2DicomConverter.cpp#L58-L62.

Good catch, i missed the dimension assignment. The `MultiFrameDimensionModule` can stay like it is (first dimension `Referenced Segment Number`, second `Image Position Patient`) but the dimension assignment in the `FrameContentMacro` has to be updated as well.

After thinking about it, I assume it would be enough to also update the `(Referenced) Segment Number` used as 1st dimension value in `Dimension Index Value` in `FrameContentMacro`. The selection of frames and their order (denoted by second dimension) for the display stays the same and does not change if just the `Segment Number`s are updated.

apps/seg/itkimage2segimage.xml
```diff
@@ -64,6 +64,15 @@
...
+      <name>sortByLabelID</name>
```
fedorov, line 68: How about renaming this as `useLabelIDasSegmentNumber`?

michaelonken:
> > I assumed that labels are unique in the input. If this assumption does not hold, then indeed the sorting does not make sense (since the idea to use the original labels as Segment Numbers cannot work then if two labels use the same number)
> 
> Aha! That's the key. Yes, I do not think we can assume anything about the label uniqueness. So what do we do with this PR?

As discussed (offline), I am pretty sure I know how to fix this. But we will make the flag not the default behavior for now.

apps/seg/itkimage2segimage.xml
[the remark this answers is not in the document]
fedorov: I agree! I made this comment in the beginning of my review, and if I revisited it after the review was done, I would most definitely revert it!

PR #493, opened 2024-02-13
WIP: Use DcmItem in API where possible.

### LINKED ISSUE #390: Itkimage2dcmSegmentation and itkimage2paramap should support "in memory" DCM sources.
Status:
Currently the interfaces of `Itkimage2dcmSegmentation` and `itkimage2paramap` build up a dependency to `DcmDataset` as you pass a vector of DcmDataset* representing the dicom source images.

Proposal: 
As far as I can see in the implementation of dcmqi only parts of the class interface is used that is directly inherited from `DcmItem`.
If this is correct and I am not mistaken this would mean that you could easily switch to the following signatures:

 ```c++
    static DcmDataset* itkimage2dcmSegmentation(vector<DcmItem*> dcmSourceItems,
                                                vector<ShortImageType::Pointer> segmentations,
                                                const string &metaData,
                                                bool skipEmptySlices=true);

    static DcmDataset* itkimage2paramap(const FloatImageType::Pointer &parametricMapImage, vector<DcmItem*> dcmSourceItems,
                                        const string &metaData);
```

It would not break existing code, but ensures the possibility to use the methods, when you have no dcm files but all informations you need. As it is e.g. in some of our use cases, where we have the values of the dcm tags at hand but not the original files. The change in the function signature would gurantee that one can directly populate DcmItems in memory and pass them instead of trying to get the files or "simulate" them by hacking temporary files with the needed information; because nothing DcmDataset-specific is used.

Have I missed something? What do you think.

If the change proposal is supported and there is reasoning against it, we could also provide a patch sooner then later, because it would ensure that we do not implement against implementation details that might change.

Looking forward to read your thoughts.

### PR DESCRIPTION
**Pull request by michaelonken**
This is work in progress, do NOT merge.

The API  change in dcmqi requires an updated DCMTK. 

However, if I use a recent DCMTK version in the build (by pointing the superbuild to it), I run into linker errors since the build tries to link DCMTK libraries twice (one time each lib with full path, which works fine, and another time by just the library name, which fails), see attached log ([linker.log](https://github.com/QIICR/dcmqi/files/14260822/linker.log))

@jcfr I think you authored the dcmqi suberbuild: Any idea what goes wrong? I fiddled around with it quite some time but could not find out where the "short" libs are added, or how to remove them. Maybe this is a quick one for you...

### DIFF
--- a/CMakeExternals/DCMTK.cmake
+++ b/CMakeExternals/DCMTK.cmake
@@ -28,13 +28,13 @@ if(NOT DEFINED DCMTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     ${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/commontk/DCMTK.git"
+    "${EP_GIT_PROTOCOL}://github.com/michaelonken/dcmtk.git"
     QUIET
     )
 
   ExternalProject_SetIfNotDefined(
     ${proj}_REVISION_TAG
-    "patched-DCMTK-3.6.6_20210115"
+    "use_dcmitem_instead_of_dcmdataset"
     QUIET
     )
 
@@ -94,11 +94,14 @@ if(NOT DEFINED DCMTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
       -DDCMTK_WITH_PNG:BOOL=OFF # see github issue #25
       -DDCMTK_WITH_TIFF:BOOL=OFF  # see github issue #25
       -DDCMTK_WITH_XML:BOOL=OFF  # see github issue #25
-      -DDCMTK_WITH_ICONV:BOOL=OFF  # see github issue #178
+      -DDCMTK_WITH_ICU:BOOL=OFF  # make sure to not include any ICU dependencies
+      -DDCMTK_WITH_ICONV:BOOL=OFF  # make sure to not include any iconv dependencies
       -DDCMTK_WITH_SNDFILE:BOOL=OFF # see github issue #395
       -DDCMTK_FORCE_FPIC_ON_UNIX:BOOL=ON
       -DDCMTK_OVERWRITE_WIN32_COMPILER_FLAGS:BOOL=OFF
-      -DDCMTK_ENABLE_BUILTIN_DICTIONARY:BOOL=ON
+      -DDCMTK_DEFAULT_DICT:STRING=builtin
+      -DDCMTK_ENABLE_CHARSET_CONVERSION:STRING=oficonv
+      -DDCMTK_ENABLE_BUILTIN_OFICONV_DATA:BOOL=ON
       -DDCMTK_ENABLE_PRIVATE_TAGS:BOOL=ON
       -DDCMTK_COMPILE_WIN32_MULTITHREADED_DLL:BOOL=ON
       -DDCMTK_ENABLE_STL:BOOL=ON
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,6 +274,7 @@ set(export_targets TRUE)
 if(DCMQI_SUPERBUILD_BINARY_DIR)
   string(FIND ${ITK_DIR}   "${DCMQI_SUPERBUILD_BINARY_DIR}" itk_here)
   string(FIND ${DCMTK_DIR} "${DCMQI_SUPERBUILD_BINARY_DIR}" dcmtk_here)
+  message(STATUS "---------------------------------------- DCMTK DIR: ${DCMTK_DIR} ------------------------------------------")
 
   if(NOT ${itk_here} EQUAL -1 OR NOT ${dcmtk_here} EQUAL -1)
     set(export_targets FALSE)
--- a/apps/paramaps/itkimage2paramap.cxx
+++ b/apps/paramaps/itkimage2paramap.cxx
@@ -45,7 +45,8 @@ int main(int argc, char *argv[])
                         (std::istreambuf_iterator<char>()));
 
   try {
-    DcmDataset* result = dcmqi::ParaMapConverter::itkimage2paramap(parametricMapImage, dcmDatasets, metadata);
+    std::vector<DcmItem*> toItems(dcmDatasets.begin(), dcmDatasets.end());
+    DcmDataset* result = dcmqi::ParaMapConverter::itkimage2paramap(parametricMapImage, toItems, metadata);
 
     if (result == NULL) {
       std::cerr << "ERROR: Conversion failed." << std::endl;
--- a/apps/seg/itkimage2segimage.cxx
+++ b/apps/seg/itkimage2segimage.cxx
@@ -120,9 +120,10 @@ int main(int argc, char *argv[])
     }
     segmentations = segmentationsReordered;
   }
-
+  // Copy to DcmItem* vector to use it in itkimage2dcmSegmentation() call
+  std::vector<DcmItem*> dcmItems(dcmDatasets.begin(), dcmDatasets.end());
   try {
-    DcmDataset* result = dcmqi::Itk2DicomConverter::itkimage2dcmSegmentation(dcmDatasets,
+    DcmDataset* result = dcmqi::Itk2DicomConverter::itkimage2dcmSegmentation(dcmItems,
                                                                              segmentations,
                                                                              metadata,
                                                                              skipEmptySlices,
--- a/include/dcmqi/ConverterBase.h
+++ b/include/dcmqi/ConverterBase.h
@@ -298,8 +298,8 @@ namespace dcmqi {
     }
 
     // AF: I could not quickly figure out how to template this function over image type - suggestions are welcomed!
-    static vector<vector<int> > getSliceMapForSegmentation2DerivationImage(const vector<DcmDataset*> dcmDatasets,
-                                                                                       const ShortImageType::Pointer &labelImage) {
+    static vector<vector<int> > getSliceMapForSegmentation2DerivationImage(const vector<DcmItem*> dcmDatasets,
+                                                                           const ShortImageType::Pointer &labelImage) {
       // Find mapping from the segmentation slice number to the derivation image
       // Assume that orientation of the segmentation is the same as the source series
       unsigned numLabelSlices = labelImage->GetLargestPossibleRegion().GetSize()[2];
--- a/include/dcmqi/Dicom2ItkConverter.h
+++ b/include/dcmqi/Dicom2ItkConverter.h
@@ -97,7 +97,7 @@ protected:
      *
      * @param segDataset A pointer to the DICOM dataset containing the metadata.
      */
-    void populateMetaInformationFromDICOM(DcmDataset* segDataset);
+    void populateMetaInformationFromDICOM(DcmItem* segDataset);
 
     /**
      * Helper method that uses the OverlapUtil class to retrieve non-overlapping segment
--- a/include/dcmqi/Itk2DicomConverter.h
+++ b/include/dcmqi/Itk2DicomConverter.h
@@ -39,7 +39,7 @@ namespace dcmqi {
     /**
      * @brief Converts itk images data into a DICOM Segmentation object.
      *
-     * @param dcmDatasets A vector of DICOM datasets with the images that the segmentation is based on.
+     * @param dcmDatasets A vector of DICOM items (usually dataset level) with the images that the segmentation is based on.
      * @param segmentations A vector of itk images to be converted.
      * @param metaData A string containing the metadata to be used for the DICOM Segmentation object.
      * @param skipEmptySlices A boolean indicating whether to skip empty slices during the conversion.
@@ -59,7 +59,7 @@ namespace dcmqi {
      *        display.
      * @return A pointer to the resulting DICOM Segmentation object.
      */
-    static DcmDataset* itkimage2dcmSegmentation(vector<DcmDataset*> dcmDatasets,
+    static DcmDataset* itkimage2dcmSegmentation(vector<DcmItem*> dcmDatasets,
                           vector<ShortImageType::Pointer> segmentations,
                           const string &metaData,
                           bool skipEmptySlices=true,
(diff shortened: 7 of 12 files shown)

### CODE REVIEW

**Comment by fedorov** [comment]
@michaelonken we have been using a patched version of DCMTK: https://github.com/commontk/DCMTK/commits/patched-DCMTK-3.6.6_20210115. I do not know what is the issue, but maybe one of those patches is needed?

**Comment by michaelonken** [comment]
Thank you Andrey. As far as I can see the patches are just two backported features.

My guess is that it has to do with changes in DCMTKTargets.cmake (in DCMTK). And/Or that somehow the expectations in the ITK build and the dcmqi build are different (the latter one is failing when linking its apps), and the superbuild settings are not forwarded into both projects the same way.

**Comment by michaelonken** [comment]
I got this.

Probably fixed with dfae36d1ab896fad294caa148a92b9864061280a. DCMTK changed its exported targets a bit, mostly moving them into namespace DCMTK. Now I ask CMake to link dcmqi and ITK to DCMTK::DCMTK which will automatically refer to all exported DCMTK libraries.

**Comment by michaelonken** [comment]
Ok, here is how one could fix it:

Earlier I already changed dcmqi/libsrc/CMakeLists.txt to replace
`set(_dcmtk_libs ${DCMTK_LIBRARIES})` with
`set(_dcmtk_libs DCMTK::DCMTK)` 

The change to DCMTK::DCMTK has to do with how DCMTK changed its exports in DCMTKTargets.cmake to a namespaced version (`DCMTK::...`) which also offers DCMTK::DCMTK for conveniently adding all DCMTK lib targets at once. (Maybe the old version with `${DCMTK_LIBRARIES})`  also can be fixed, but why not use DCMTK::DCMTK if its available.)

However, during superbuild, dcmqi also downloads a specific ITK version: 
`https;//github.com/Slicer/ITK.git, commit be914a`
That ITK copy also uses DCMTK.

The ITK version downloaded uses in `Modules/ThirdParty/DCMTK/CMakeLists.txt`:
`set(ITKDCMTK_LIBRARIES ${DCMTK_LIBRARIES})`

This would need to be changed to:
`set(ITKDCMTK_LIBRARIES DCMTK:::DCMTK)`
and then dcmqi successfully builds.

I don't understand the full mechanics of the dcmqi superbuild, so I am not sure what to make out of this:

- Can we fix this in ITK?
- Or can/should we patch the downloaded ITK to use DCMTK::DCMTK?
- And/Or, can/should the downloaded ITK inherit the linked libraries from dcmqi (I favor DCMTK::DCMTK)?
- Is there an easier fix?

**Comment by fedorov** [comment]
Michael, I believe the idea one idea behind the superbuild is to support the situation where dcmqi is built using ITK/DCMTK and other dependencies that are already available outside of dcmqi suiperbuld (which is the case when it is built as a Slicer extension). I do not know for sure, but maybe that somehow contributes to figuring out the solution.


Have you tried building with new DCMTK by pointing to it in the cmake variable as we do here: https://github.com/QIICR/dcmqi/blob/master/.github/workflows/cmake-win.yml#L60 ?

**Comment by michaelonken** [comment]
Thanks Andrey. 

The problem at the core seems to be that ITK uses ${DCMTK_LIBRARIES} which is the list of _library names_ (ofstd, dcmdata,...). ITK as such build and links fine also with a recent DCMTK.

DCMQI then links against ITK and DCMTK, plus as I understand it inherits the library requirements from ITK in ${ITK_LIBRARIES}. The latter still contains a list of DCMTK library names resulting in linker flags -lofstd -ldcmdata and so on. But, DCMQI build does not seem to specify the library path where to find these libraries. This worked for some reason for older DCMTK versions probably due to changed variables in DCMTKTargets/Config.cmake .

I can and want to change DCMQI linkage to link to the newly DCMTK-defined target DCMTK::DCMTK which then links against all DCMTK libraries, using their full library path, i.e. -l/home/test/dcmtk-build/lib/ofstd.a . However, the linker still adds additionally -lofstd -ldcmdata etc. I tried a long time to remove that part from the DCMQI linker command inherited by ${ITK_LIBRARIES}. I tried to remove the pure library names from the linker command by removing them from ${ITK_LIBRARIES} in DCMQI CMake files but without effect; or I removed it too early or too late in the CMake process.

So, if someone knows how to remove these effectively from the DCMQI linker call let me know.

Another solution is to patch the downloaded ITK to also link against DCMTK::DCMTK if it finds a recent DCMTK but that's probably even more hackish (but works if I do this manually). Or wait that DCMTK::DCMTK or another mechanism goes into ITK some day.

An update to a recent DCMTK version is required by this PR but also for the desired feature to not fail writing segmentation objects etc. when some attributes have invalid values. Also I think it makes sense to update DCMTK more often, since the version in DCMQI/Slicer is quite old.

**Comment by fedorov** [comment]
I agree, it is important to figure this out - both for dcmqi and Slicer! I started a discussion in Slicer forum: https://discourse.slicer.org/t/slicer-dcmtk-needs-an-update/34363.

**Comment by fedorov** [comment]
Relevant work from @jamesobutler in https://github.com/Slicer/Slicer/pull/6709.

**Comment by thewtex** [comment]
We are able to build DCMQI against the latest DCMTK. However, we may want to expose more DCMTK modules in what ITK exposes if people want to use ITK's DCMTK for DCMQI. In the long run, we should find an effort that would enable improvement of ITK's use of CMake targets so they can be used in a more fine-grained way.

**Comment by fedorov** [comment]
> We are able to build DCMQI against the latest DCMTK.

I rebased Michael's branch, but it is still failing...

**Comment by fedorov** [comment]
@michaelonken it is still failing, but it is not a link error anymore

```
/Applications/Xcode_15.0.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -DDCMTK_BUILD_IN_PROGRESS -DDCMTK_ENABLE_BUILTIN_OFICONV_DATA -DUSE_NULL_SAFE_OFSTRING -D_BSD_COMPAT -D_BSD_SOURCE -D_OSF_SOURCE -D_REENTRANT -D_XOPEN_SOURCE_EXTENDED -Dofstd_EXPORTS -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK-build/config/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/ofstd/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/oflog/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmdata/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmimgle/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmimage/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmjpeg/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmjpls/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmtls/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmnet/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmsr/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmsign/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmwlm/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmqrdb/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmpstat/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmrt/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmiod/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmfg/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmseg/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmtract/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmpmap/include -I/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/dcmect/include -I/include -D_XOPEN_SOURCE_EXTENDED -D_BSD_SOURCE -D_BSD_COMPAT -D_OSF_SOURCE -D_DARWIN_C_SOURCE -fPIC -g -DDEBUG -std=c++14 -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk -mmacosx-version-min=14.0 -Wall -MD -MT ofstd/libsrc/CMakeFiles/ofstd.dir/ofchrenc.cc.o -MF ofstd/libsrc/CMakeFiles/ofstd.dir/ofchrenc.cc.o.d -o ofstd/libsrc/CMakeFiles/ofstd.dir/ofchrenc.cc.o -c /Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/ofstd/libsrc/ofchrenc.cc
/Users/runner/work/dcmqi/dcmqi/dcmqi-build/DCMTK/ofstd/libsrc/ofchrenc.cc:299:10: fatal error: 'dcmtk/oficonv/iconv.h' file not found
#include "dcmtk/oficonv/iconv.h"
         ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

michaelonken:
I think the main issue is still the linker. This is what I see:

- Linux:  Linker error, DCMTK libraries not found (path to DCMTK libs wrong?).
- Windows: Linker error, DCMTK libraries not found (path to DCMTK libs wrong?)
- Mac: This is different. oficonv.h is not found but other DCMTK includes before are. I suppose it is a cache issue but did not look deeper so far.  I experience see the same error if I still have an old version of DCMTK in the dcmqi build tree.

On the linker error on Linux and Windows:

I think this is still the error I saw earlier when trying to link dcmqi to a newer DCMTK. I can see the same behaviour on my own Linux box. The following seems to happen: 

`dcmqi/libsrc/CMakelists.txt` (at least, later more) adds its "own" DCMTK libraries (via  ${_dcmtk_libs}) here:

`target_link_libraries(${lib_name} PUBLIC
  ${_dcmtk_libs}
  ${ITK_LIBRARIES}
  $<$<NOT:$<BOOL:${DCMQI_BUILTIN_JSONCPP}>>:${JsonCpp_LIBRARY}>
  )
`
  ${_dcmtk_libs} will contain the full paths to dcmtk libs, e.g. /home/michael/b/dcmqi_linux/DCMTK-build/lib/libdcmnet.a on my system. These are also found by the linker.

However, the call above also links to   ${ITK_LIBRARIES} which contain the DCMTK libraries again, "imported" from ITK, e.g. "dcmnet;dcmdata;.." and so on, which will not be found later, since there is no related library path set.

So there are three possibilities I think:
1. Remove plain list of DCMTK libs (-ldcmdata -ldcmnet...) from list of link libraries.
2. Add link directory where those libraries reside so linker can find them.
3. Remove dcmqi copy of DCMTK libss (-lhome/michael/b/dcmqi_linux/DCMTK-build/lib/libdcmnet.a..) from the call.

ad 1)  I tried to remove DCMTK libs from   ${ITK_LIBRARIES} in dcmqi/libsrc/CMakeLists.txt via
`list(REMOVE_ITEM ITK_LIBRARIES ofstd oflog oficonv`
but it does not help.

So my guess is that   ${ITK_LIBRARIES} are added somewhere else as well. Even if remove DCMTK libs from main dcmqi/CMakeLists.txt file the "plain" list of DCMTK libs (-ldcmdata -ldcmnet...) still is being used by the linker. So *somewhere* in the CMake code there must be a possibility to remove DCMTK from that ITK-imported library list (?).

ad 2) I did not try this, since have bad feeling of linking against two copies of DCMTK libs. We could try to point this directory to the dcmqi copy of DCMTK libs but that would be  a workaround; we should make sure every DCMTK lib only shows up once in the linker library list.

ad 3) Add the library dir where the ITK copy of DCMTK libs reside. Since I don't know in which CMake variable this directory could be found, I did not give it try yet.

I assume the preferred solution would be 1. since this would allow us to link against a newer DCMTK in dcmqi independent of what ITK uses (only one explicitly defines DCMTK_DIR or so to point to the ITK copy of DCMTK).

fedorov:
Thank you @michaelonken. 

@jcfr pinging you here since we discussed DCMTK update in Slicer yesterday.

@thewtex I tried to build dcmqi against the latest DCMTK 3.6.8, and it is failing - see #500.

thewtex:
After @jadh4v has finished the seg and parameter map integration in ITK-Wasm, we should update the dcmtk libraries built and provided by ITK to be the union of what is there now, what dcmqi needs, and what ITK-Wasm needs.

jcfr:
Working on updating DCMTK in Slicer & CTK, will the we have to also backport the changes from https://github.com/michaelonken/dcmtk/commits/use_dcmitem_instead_of_dcmdataset/ ?

Ideally, those should be integrated into DCMTK proper.

michaelonken:
Thanks for the reminder, I will merge it in this week.

jcfr:
> Thanks for the reminder, I will merge it in this week.

Great :pray: 

Once those are merged, we should have integrated the following pull request and it will then be trivial to update DCMTK:
* https://github.com/Slicer/Slicer/pull/8018

> [!IMPORTANT]
> We will wait for your changes before initiating the `Slicer 5.8` release process :rocket:

fedorov:
> Thanks for the reminder, I will merge it in this week.

Sorry, I am confused. Those would be merged upstream into DCMTK, but I thought we wanted to update Slicer/dcmqi to the latest published DCMTK release 3.6.8. Shouldn't we treat those two separately? I would think that update to 3.6.8 should be ready to go.

michaelonken:
The original patch to DCMTK that is required by this issue/patch to dcmqi is now in DCMTK and will show up on master within a few days. We are currently planning for a DCMTK release before the end of the year.

fedorov:
blocked by https://github.com/QIICR/dcmqi/pull/500

rfloca:
@michaelonken: Please excuse the delay. Yes, https://github.com/QIICR/dcmqi/commit/09e1833e24d07a9f3eeba6c07a8442a1ce39dd66 containes the API change desired.

I will also make a PR for an adaption of itkimage2dcmSegmentation on top of your changes. But for now having DCMQI having this API changes and be compilable with a recent DCMTK, is the only thing we would need to take it from there.

Thank you very much for investing your time!

PR #495, opened 2024-04-22
allow saving merged segments to all formats

fedorov:
I do not see a reason to restrict this feature to NRRD


michaelonken (approved):
I'm not very familiar with other output formats, but if they allow the same "label map approach" (where each pixel value represents a specific segment), then I don't see any blockers.

PR #500, opened 2024-06-25
ENH: Update from DCMTK 3.6.6 to 3.6.8

fedorov:
Update `CMAKE_CXX_STANDARD` from 14 to 17, and simplify settings of `CMAKE_CXX_STANDARD` variable based of Slicer/Slicer@d762c0465e (`"COMP: Configure
DCMTK with CMAKE_CXX_STANDARD"`, 2022-02-21)

Update ITK version to match Slicer ones

Fix macOS support simplifying CMakeOXVariables module based of the following
commits:
* Slicer/Slicer@f60a86f4ae (`"ENH: Use CMake 3.0 default method to find OSX SDK root"`, 2019-07-02)
* Slicer/Slicer@ac191b40f7 (`"ENH: Simplify OSX Deployment target setting"`, 2022-02-16)

In DCMTK and ITK external projects, respectively remove reference to unused CMake variables `ep_common_cache_args` and `ep_project_include_arg`.

Remove the `DCMTK_ENABLE_BUILTIN_DICTIONARY` option, obsolete since commit DCMTK/dcmtk@4ab084d3f ("Harmonize dictionary options.", 2021-09-01), and set `DCMTK_DEFAULT_DICT` instead.

Introduce `EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS` CMake variable to pass CMake arguments.


fedorov:
@jcfr thank you for looking into this!

Regarding windows, in the past I used Appveyor to build ITK and DCMTK dependencies in order to fit the total build time within limits that were in place at that time. I tried switching to build all dependencies as part of the overall workflow, but that unfortunately failed with "error MSB8066", which I suspect was because of the path length exceeding allowance. I then opened a separate PR where I also reduced the directory name a little: https://github.com/QIICR/dcmqi/pull/516. I also tried to update the Appveyor project to make package, but Appveyor fails without clear indication why (this is not entirely surprising, since last time I used it was 2 years ago when DCMTK was updated last time!).

I spent couple of hours on this today, but didn't find a solution.

Can you for now focus on mac, where tests are still failing?

fedorov:
@jcfr after looking a bit more, I realized that the external dependencies were downloaded but not used in the workflow.

As you can see in the windows build for #516, it completes correctly, so the build error in this PR is a new regression. Perhaps this is due to the path length exceeding the limit. Reducing the top-level folder to single character didn't fix the problem.

I merged #516 and rebased this PR to get rid of that misleading portion of the workflow.

jcfr:
Thanks for working on this. I am now headed out in the wilderness for a few days and will be able to have a look when back. 🙏 thanks for the patience.

michaelonken:
Note that DCMTK 3.6.9 has been released 3 months ago. I there any reason from your side to stick to 3.6.8 for now (which was current when this pull request has been opened)?

Of course, if you need any information about 3.6.9 or DCMTK in general, let me know.

jcfr:
> Is there any reason from your side to stick to 3.6.8 for now (which was current when this pull request has been opened)?

Thanks for checking. We are matching the version currently used in Slicer. Once this PR is finalized and integrated, we will look into transitioning both to 3.6.9.

jcfr:
@fedorov Once the CI completes and is green, this will be ready for integration using either "Merge" or "Rebase & Merge".

🙏 Thanks again for the patience.

fedorov:
Thank you @jcfr, looks promising!

@rfloca fyi - hopefully this will be addressed soon and will allow proceed with the issue that is holding you up.

jcfr:
Turns out that ITK is already built with `BUILD_TESTING` to `OFF`. In this particular case, we would need to explicitly disable the `ITKTestKernel` module.

Before doing so, we updated to the version ITK to match the one recently updated in Slicer through the following pull request:
* https://github.com/Slicer/Slicer/pull/8339

fedorov:
Do you know why this issue is not happening in Slicer superbuild?

jcfr:
> Do you know why this issue is not happening in Slicer superbuild?

This was likely related to the fact we are building the libraries statically when building dcmqi standalone.

Considering that updating ITK address this, there is not need to explicitly disable the `ITKTestKernel` module.

jcfr:
This is ready to be integrated using either `Merge` or `Rebase & Merge`

jcfr:
Once this is integrated, we will likely have to also fix the "publish" step that is now using an outdated approach.

fedorov:
JC, thank you, so great to see this ready! I pinged @michaelonken in case he would like to take a look.

> Once this is integrated, we will likely have to also fix the "publish" step that is now using an outdated approach.

This is probably why the latest release workflow didn't route the linux and mac binaries correctly.

image

jcfr:
Once this is integrated, we will on updating dcmth to 3.6.9 in both Slicer and DCMQI.

Also, if you grant me admin access (at least for the next few weeks) to dcmqi I will be able to more easily address the upload of release packages.

jcfr:
> admin access granted!

:pray: This will be helpful to finalize

PR #502, opened 2024-06-28
Use const pointer to input segmentations

jadh4v:
Use constant pointers to input segmentation images (`itk::Image::ConstPointer`) since we don't expect the function to change these images.


fedorov (changes requested):
CI is red!

fedorov (approved):

PR #511, opened 2024-12-10
COMP: Update python executable location and usage

thewtex:
Use the CMake Python3 module instead of PythonInterp, which it has
replaced with the Interpreter COMPONENT.

Use Python 3 instead of Python 2, which is end-of-life.

Ensure we use ${Python3_EXECUTABLE} in the test commands.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -22,7 +22,7 @@ jobs:
         fetch-tags: true 
     - uses: actions/setup-python@v4
       with:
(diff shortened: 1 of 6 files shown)


fedorov:
Makes sense, thank you @thewtex! 👍

fedorov:
Should I make a new dcmqi release?

thewtex:
@fedorov thanks for the review and merge!

I will just use the Git hash, so a new release is not required for my needs.

PR #517, opened 2025-05-30
Bump DCMTK version to 3.6.9 and import DcmItem.

michaelonken:
Bump DCMTK version to latest DCMTK release 3.6.9.

This also allows to import dataset references from DcmItem instead of DcmDataset which is easier to use when importing from memory.

Put this into a pull request to build run the tests on various platforms before merging into master.

### DIFF
--- a/CMakeExternals/DCMTK.cmake
+++ b/CMakeExternals/DCMTK.cmake
@@ -41,13 +41,13 @@ if(NOT DEFINED DCMTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     ${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/DCMTK/dcmtk.git"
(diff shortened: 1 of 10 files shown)


fedorov:
@michaelonken note that in the end dcmqi should work with Slicer, and if we need to have 3.8.9 in Slicer, we need to update it there too. As you can see from https://github.com/commontk/DCMTK/tree/patched-DCMTK-3.6.8_20241024, it has this additional patch https://github.com/commontk/DCMTK/commit/ea07125078cd097245867a71d8fba8b36fd92878. Maybe it is already handled in 3.8.9 upstream?

@jcfr is there anything else? Should we work on updating DCMTK in Slicer to 3.8.9?

michaelonken:
Hi,

yes, I know this patch. 

I tried it today on the latest DCMTK master but since it didn't work (and I think it went too far with what it tried to do).

I isolated the actually required part required for external projects (adapting osconfig.h) and pushed it to my michaelonken/dcmtk repo, and make dcmqi use it in this PR for testing (before pushing it to official DCMTK master). The Windows build is still running but I'm confident ... :-)

fedorov:
I think it is better to freeze version in Slicer. I will do that!

PR #534, opened 2026-04-20
Add segment for background if required.

michaelonken:
When converting DICOM binary segmentations to DICOM labelmaps, some pixels are likely not covered from any of the the segments. Since the pixels are initially set to 0, all non-covered pixels will be "background" and must be marked as such.

This is done by scanning all pixels after conversion and if a 0 is found, a segment using codes for denoting "background" will be inserted automatically.

The alternative would be to set Pixel Padding Value. This could be introduced as an option later, but for now the background segment seems to be better supported by consuming applications.


fedorov:
Superseded by #535

PR #535, opened 2026-05-08
Ft upgrade dcmtk

michaelonken:
See [commit message](https://github.com/QIICR/dcmqi/pull/535/changes/08ece479a8962b355c868820f61a09c590e0e7d4).

This also includes/duplicates changes from PR #534.

PR #536, opened 2026-05-08
Add direct ITK to DICOM Labelmap conversion

michaelonken:
Add direct ITK to DICOM Labelmap conversion.

Also includes PRs #534 and #535 (the latter being probably optional).

Extend itkimage2segimage with a --segmentationType option (default
"binary", preserving prior behavior) that, when set to "labelmap",
emits a DICOM Labelmap Segmentation (Sup 243) directly from an ITK
image. Previously the only way to produce a labelmap was to convert
to a binary Segmentation first and then run bin2labelsegimage;
this removes the intermediate step.

In labelmap mode foreground segments are numbered 1..N (or taken
from labelIDs when --useLabelIDAsSegmentNumber is set); pixel value
0 is reserved for the implicit background. If any frame contains
pixel 0 a Background segment (Property Type DCM 125040) is added
automatically so every pixel value is covered by a Segment Sequence
item, as required by Sup 243. Bit depth is 8-bit when the segment
count fits, else 16-bit. Overlapping foreground segments are
rejected because each labelmap pixel can carry only one segment
number.

The frame-scan + background-segment-creation logic that this new
path needs is identical to what Bin2Label already does for
binary->labelmap conversion. It is consolidated into a shared
ConverterBase::addBackgroundSegmentIfNeeded() helper, and
DcmBinToLabelConverter's member function becomes a thin wrapper
that delegates and adds the PALETTE-mode LUT prepend on top.

Also adds a 2-segment metadata example
(doc/examples/seg-example_liver_spine.json) used by new CTest
entries:
- itkimage2segimage_makeSEG_labelmap: single-segment conversion.
- itkimage2segimage_makeSEG_labelmap_multi: multi-segment
  conversion. Regression guard for segment numbering, which is
  not exercised by single-segment inputs.
- itkimage2segimage_makeSEG_labelmap_multi_labelID: same with
  --useLabelIDAsSegmentNumber.
- itkimage2segimage_makeSEG_labelmap_overlap: liver/heart pair
  overlaps on 522 pixels; conversion must fail (WILL_FAIL).
- segimage2itkimage_makeNRRD_from_labelmap_{direct,multi}:
  pixel-perfect round-trip back to NRRD against existing baselines.

dciodvfy is intentionally not run on the new labelmap outputs;
all labelmap segmentations emitted by dcmqi today (both Bin2Label
and this new path) fail dciodvfy because Type 2C PatientOrientation
is not populated under GeneralImage. This is a pre-existing
gap and is documented inline in the CMakeLists.

### DIFF
--- a/CMakeExternals/DCMTK.cmake
+++ b/CMakeExternals/DCMTK.cmake
@@ -47,10 +47,12 @@ if(NOT DEFINED DCMTK_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     ${proj}_GIT_TAG
-    # DCMTK 3.7.0 ++ (December 2025).
+    # DCMTK 3.7.0 ++ (May 8th, 2026).
     # This is DCMTK version with some extra fixes on segmentations and parametric maps,
-    # which have been released a few days after DCMTK 3.7.0.
-    "3e85b37444107e93550167c2284e64b4881b0fcb"
+    # which have been released a few days after DCMTK 3.7.0, as well as performance
+    # optimizations for segmentation/parametric map. It also includes a simpler
+    # mechanism to add external modules to the DCMTK CMake build.
+    "2dd54ca1c28100b820ccf1383a0948e889246f96"
     QUIET
     )
 
--- a/apps/seg/Testing/CMakeLists.txt
+++ b/apps/seg/Testing/CMakeLists.txt
@@ -30,6 +30,68 @@ dcmqi_add_test(
     --outputDICOM ${MODULE_TEMP_DIR}/liver.dcm
   )
 
+dcmqi_add_test(
+  NAME ${itk2dcm}_makeSEG_labelmap
+  MODULE_NAME ${MODULE_NAME}
+  COMMAND $<TARGET_FILE:${itk2dcm}>
+    --inputMetadata ${CMAKE_SOURCE_DIR}/doc/examples/seg-example.json
+    --inputImageList ${BASELINE}/liver_seg.nrrd
+    --inputDICOMDirectory ${DICOM_DIR}
+    --outputDICOM ${MODULE_TEMP_DIR}/liver-labelmap-direct.dcm
+    --segmentationType labelmap
+  )
+
+# Multi-segment labelmap from two non-overlapping single-segment ITK inputs.
+# Verifies that segment numbering for foreground starts at 1 (regression guard
+# for the bug where the first foreground segment got number 0 and collided
+# with the implicit background pixel value).
+dcmqi_add_test(
+  NAME ${itk2dcm}_makeSEG_labelmap_multi
+  MODULE_NAME ${MODULE_NAME}
+  COMMAND $<TARGET_FILE:${itk2dcm}>
+    --inputMetadata ${CMAKE_SOURCE_DIR}/doc/examples/seg-example_liver_spine.json
+    --inputImageList ${BASELINE}/liver_seg.nrrd,${BASELINE}/spine_seg.nrrd
+    --inputDICOMList ${DICOM_DIR}/01.dcm,${DICOM_DIR}/02.dcm,${DICOM_DIR}/03.dcm
+    --outputDICOM ${MODULE_TEMP_DIR}/liver_spine-labelmap.dcm
+    --segmentationType labelmap
+  )
+
+# Same as above but exercises --useLabelIDAsSegmentNumber together with
+# --segmentationType labelmap.
+dcmqi_add_test(
+  NAME ${itk2dcm}_makeSEG_labelmap_multi_labelID
+  MODULE_NAME ${MODULE_NAME}
+  COMMAND $<TARGET_FILE:${itk2dcm}>
+    --inputMetadata ${CMAKE_SOURCE_DIR}/doc/examples/seg-example_liver_spine.json
+    --inputImageList ${BASELINE}/liver_seg.nrrd,${BASELINE}/spine_seg.nrrd
+    --inputDICOMList ${DICOM_DIR}/01.dcm,${DICOM_DIR}/02.dcm,${DICOM_DIR}/03.dcm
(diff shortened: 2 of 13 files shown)


fedorov (commented):
## Review: Direct ITK-to-DICOM Labelmap Segmentation

**+743/-45 lines | 13 files changed**

### Overview

Adds a `--segmentationType labelmap` option to `itkimage2segimage`, enabling direct ITK→DICOM Labelmap Segmentation (Sup 243) conversion without the previous two-step binary→labelmap pipeline. Consolidates shared background-segment creation logic into `ConverterBase::addBackgroundSegmentIfNeeded()`, fixes a stale `fgder` deletion guard, improves memory safety by replacing `new Uint8[]` with `std::vector`, removes debug prints, and bumps the DCMTK dependency to a May 2026 commit.

---

### Positives

- **Refactoring is well-designed.** Extracting the background-segment logic into `ConverterBase` is clean; `DcmBinToLabelConverter::addBackgroundSegmentIfNeeded()` becomes a thin wrapper adding only the PALETTE-specific LUT prepend.
- **Test coverage is thorough:** single-segment, multi-segment, `--useLabelIDAsSegmentNumber`, expected-failure overlap, and pixel-perfect NRRD round-trips.
- **Memory safety improvement.** `Uint8 *frameData = new Uint8[frameSize]` is now a `std::vector<Uint8>` — both for the binary path (fix) and the new labelmap path (clean from the start).
- **Correct derivation-image pre-computation.** The new `slice2derimgPerFile` vector computes geometry maps once before the slice loop, avoiding repeated work per segment file.
- **`segmentsOverlap = "NO"` for labelmap output** is correct since the converter explicitly rejects overlapping segments.
- **`fgder` deletion fix** (removing `if(referencesGeometryCheck)` guard) is correct — `fgder` is always allocated.
- **Documentation is solid.** Docstrings cite Sup 243 section numbers; the dciodvfy gap is explained inline with a clear TODO path.

---

### Issues and Concerns

#### 1. Wrong `<channel>` in the XML descriptor (`itkimage2segimage.xml`)

```xml
<channel>output</channel>
```

`segmentationType` controls conversion *input* behavior. Should almost certainly be `<channel>input</channel>`, like the other control flags (`useLabelIDAsSegmentNumber`, `compress`, etc.). As written, Slicer module auto-generation may misclassify this parameter.

#### 2. Memory leak on `CHECK_COND` early exits inside the labelmap block (`Itk2DicomConverter.cpp`)

`CHECK_COND` expands to `return NULL` on failure. `fgppp`, `fgfc`, and `fgder` are allocated before the labelmap block and deleted at the end, but `CHECK_COND` calls like `fgfc->setStackID(...)`, `fgfc->setInStackPositionNumber(...)`, etc., can bypass those deletes. The `releaseFGs` lambda handles the two explicit error paths, but `CHECK_COND` bypasses it entirely. This is a pre-existing pattern in the binary path too, but the new path adds substantially more `CHECK_COND` calls. Consider replacing the raw pointers with `std::unique_ptr` or wrapping them in RAII.

#### 3. `CIELabColor::prepend()` assumes nothrow `new` (`Bin2Label.h`)

```cpp
Uint16* newL = new Uint16[newSize];
Uint16* newA = new Uint16[newSize];
Uint16* newB = new Uint16[newSize];
if (!newL || !newA || !newB)  // null-checks after standard new
```

Standard `operator new` throws `std::bad_alloc` on failure rather than returning null. The null guard is only effective if DCMTK is built with a nothrow allocator. Should use `new (std::nothrow)` explicitly, or restructure with `std::vector`.

#### 4. Unconditional `DCM_RecommendedDisplayCIELabValue` rule deletion in `copySegments` (`Bin2Label.cpp`)

The old code deleted this rule only in PALETTE mode. The new code deletes it unconditionally. If this is intentional for MONOCHROME2 labelmap output too (e.g., "Sup 243 prohibits per-segment CIELab in labelmap mode regardless of color model"), a brief comment would prevent future readers from reverting it.

#### 5. Pre-existing dciodvfy compliance gap (documented but worth a separate issue)

The `Testing/CMakeLists.txt` comment clearly documents that all labelmap output fails dciodvfy validation because `PatientOrientation` is not populated in `GeneralImage`. This should be tracked as a separate issue so it doesn't get lost.

---

### Summary

Solid PR overall. The architecture is clean, tests are comprehensive, and the DICOM spec compliance reasoning is well-documented. The `<channel>output</channel>` bug in the XML (issue 1) is the most impactful issue — it affects Slicer module auto-generation. The memory safety concern (issue 2) is worth addressing before merge. Items 3 and 4 are lower priority but worth a comment or minor fix.

fedorov:
@michaelonken I think this one (pre-existing) is incorrect as well - should be input!

https://github.com/QIICR/dcmqi/blob/8b161f98d2c94f8b1973a328e85289138daf6d5e/apps/seg/itkimage2segimage.xml#L85-L92

michaelonken:
> @michaelonken I think this one (pre-existing) is incorrect as well - should be input!
> 
> https://github.com/QIICR/dcmqi/blob/8b161f98d2c94f8b1973a328e85289138daf6d5e/apps/seg/itkimage2segimage.xml#L85-L92

yes, its actually an input flag; since it controls the checking of the output I accidentally selected output. 

I must check this issue though:

4. Unconditional DCM_RecommendedDisplayCIELabValue rule deletion in copySegments (Bin2Label.cpp)

michaelonken:
I think its debatable whether the --noDicomValueChecks is an input or output option. Since the tool (itk2segimage) writes DICOM as an output and the the option allows skipping checks while writing the output, I would prefer keeping it as an output option.

fedorov:
> I think its debatable whether the --noDicomValueChecks is an input or output option. Since the tool (itk2segimage) writes DICOM as an output and the the option allows skipping checks while writing the output, I would prefer keeping it as an output option.

@pieper can you help us resolve this - I believe in Slicer SEM any parameter that is passed to a CLI should be "input" channel, right?

pieper:
Yes, by Slicer CLI semantics it's a `input` parameter.  `Outputs` from the CLI would be things that Slicer reads back in, not parameters that define how the output is written.

michaelonken:
Got it 👍

PR #538, opened 2026-05-28
Add in-memory handler overload of itkimage2dcmSegmentation.

rfloca:
## Summary

  Adds a new template overload of `Itk2DicomConverter::itkimage2dcmSegmentation`
  that takes a `JSONSegmentationMetaInformationHandler&` instead of a JSON
  string. The handler overload becomes the load-bearing implementation; the
  existing string overload is reduced to a thin wrapper that parses its JSON
  argument into a handler and delegates. Behaviour for current callers is
  unchanged.

  ## Motivation

  Consumers (e.g. MITK) build their segmentation metadata programmatically
  and currently have to serialise it to JSON just to have dcmqi parse it
  back into a handler. The new overload removes that round-trip and avoids
  the brittleness of round-tripping rich metadata through JSON.

  ## Testing

  Adds `apps/seg/Testing/Itk2DicomHandlerOverloadEquivalenceTest`, which
  exercises both overloads against the same input and asserts the resulting
  datasets agree on:

  - segment sequence cardinality,
  - per-frame functional groups cardinality,
  - segment labels,
  - `SegmentsOverlap`,
  - patient/study/series context tags.

  Byte-level equality is not asserted, because DCMTK regenerates UIDs and
  timestamps on every write.

  ## Test plan

  - [ ] CI green on all platforms
  - [ ] `Itk2DicomHandlerOverloadEquivalenceTest` passes locally
  - [ ] Existing seg conversion tests still pass (string overload behaviour unchanged)

  ## Files for reviewers

  - **`include/dcmqi/Itk2DicomConverter.h`** — declares the new template overload
    `itkimage2dcmSegmentation(..., JSONSegmentationMetaInformationHandler& metaInfo, ...)`
    alongside the existing string overload. Adds a forward declaration of
    `JSONSegmentationMetaInformationHandler` so the public header stays light;
    callers using the handler overload include
    `JSONSegmentationMetaInformationHandler.h` themselves.
  - **`libsrc/Itk2DicomConverter.cpp`** — moves the full conversion body into the
    handler overload. The string overload is reduced to a few lines that parse the
    JSON into a handler and delegate. Adds explicit template instantiations of the
    new overload for the two pixel adapters already instantiated for the string
    overload (`ShortImageType` and `VectorImageAdapter`), so the ABI surface
    mirrors the existing one exactly.
  - **`apps/seg/Testing/Itk2DicomHandlerOverloadEquivalenceTest.cxx`** (new) —
    equivalence guard: runs both overloads on the same input and compares the
    resulting `DcmDataset`s on the fields listed above. This is the main thing to
    scrutinise — particularly whether the comparison set is sufficient to catch
    regressions where the two paths could silently diverge.
  - **`apps/seg/Testing/CMakeLists.txt`** — wires the new test into CTest.

  ### Suggested review focus

  1. Is the forward declaration in the public header acceptable, or would you
     prefer a full `#include` of `JSONSegmentationMetaInformationHandler.h`?
  2. Does the equivalence-test comparison set cover enough surface, or is there
     a specific field/tag you'd want asserted as well?
  3. The string overload now does an extra handler construction + `read()` step
     that previously happened inline. This is the same work as before, just
     factored — but worth a sanity check that no caller depended on side effects
     ordering.

### DIFF
--- a/apps/seg/Testing/CMakeLists.txt
+++ b/apps/seg/Testing/CMakeLists.txt
@@ -20,6 +20,31 @@ dcmqi_add_test(
   COMMAND $<TARGET_FILE:${itk2dcm}> --help
   )
 
+#-----------------------------------------------------------------------------
+# Equivalence guard for the in-memory handler overload of
+# itkimage2dcmSegmentation. The string overload is implemented as a thin
+# wrapper that parses its JSON argument into a handler and delegates to the
+# handler overload; this test asserts the two overloads produce semantically
+# equivalent DICOM datasets for the same input.
+add_executable(Itk2DicomHandlerOverloadEquivalenceTest
+  Itk2DicomHandlerOverloadEquivalenceTest.cxx)
+target_link_libraries(Itk2DicomHandlerOverloadEquivalenceTest
+  dcmqi
+  ${DCMTK_LIBRARIES})
+set_target_properties(Itk2DicomHandlerOverloadEquivalenceTest PROPERTIES
+  LABELS ${MODULE_NAME})
+
+dcmqi_add_test(
+  NAME ${itk2dcm}_handlerOverloadEquivalence
+  MODULE_NAME ${MODULE_NAME}
+  COMMAND $<TARGET_FILE:Itk2DicomHandlerOverloadEquivalenceTest>
+    ${CMAKE_SOURCE_DIR}/doc/examples/seg-example.json
+    ${BASELINE}/liver_seg.nrrd
+    ${DICOM_DIR}/01.dcm
+    ${DICOM_DIR}/02.dcm
+    ${DICOM_DIR}/03.dcm
+  )
+
 dcmqi_add_test(
   NAME ${itk2dcm}_makeSEG
   MODULE_NAME ${MODULE_NAME}
--- a/apps/seg/Testing/Itk2DicomHandlerOverloadEquivalenceTest.cxx
+++ b/apps/seg/Testing/Itk2DicomHandlerOverloadEquivalenceTest.cxx
@@ -149,7 +149,7 @@ int main(int argc, char* argv[])
   REQUIRE(resultA != nullptr);
 
   // Path B: handler overload (new).
-  dcmqi::JSONSegmentationMetaInformationHandler handler(metadata.c_str());
+  dcmqi::JSONSegmentationMetaInformationHandler handler(metadata);
   handler.read();
   std::unique_ptr<DcmDataset> resultB(
       dcmqi::Itk2DicomConverter::itkimage2dcmSegmentation(dcmDatasetsB, segmentations, handler));
--- a/apps/seg/Testing/Itk2DicomHandlerOverloadEquivalenceTest.cxx
+++ b/apps/seg/Testing/Itk2DicomHandlerOverloadEquivalenceTest.cxx
@@ -1,0 +1,188 @@
+// Equivalence test for the two itkimage2dcmSegmentation overloads.
+//
+// The handler-based overload is the load-bearing implementation; the
+// string-based overload is a thin wrapper that parses its JSON argument into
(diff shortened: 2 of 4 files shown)


michaelonken:
Hi Ralf,

as I understand the basic trick is to use the existing metadata handler directly as an argument for the conversion call, instead of the JSON string, and keep the old JSON string-based methods for backward compatibility?

P.S: There seem to be build errors, eg [here](https://github.com/QIICR/dcmqi/actions/runs/26586487786/job/78339104885?pr=538#step:9:6166), due to wrong number of arguments in CMake call.

rfloca:
Hi Michael

> as I understand the basic trick is to use the existing metadata handler directly as an argument for the conversion
> call, instead of the JSON string, and keep the old JSON string-based methods for backward compatibility?

That is correct. Our use case is e.g. that we generate all the information in memory so that we don't need to have files present on disk to write a DCM Seg. So far we populated always the handler and generated the json, which gets directly deserialzed again directly after the function call boundary. So the json is a wire format with no added value here. For the contract the handler is authorative. 

> P.S: There seem to be build errors, eg [here](https://github.com/QIICR/dcmqi/actions/runs/26586487786/job/78339104885?pr=538#step:9:6166), due to wrong number of arguments in CMake call.

Yes. My bad. I thought I had test on on my local build. 🙈  Now they are and a fix is commited. Please excuse the noise.

fedorov:
@rfloca thank you for the contribution!

Could you please rebase to `master` and update the branch? I asked Copilot to fix the failing CI cache...

rfloca:
> @rfloca thank you for the contribution!
> 
> Could you please rebase to `master` and update the branch? I asked Copilot to fix the failing CI cache...

Yes I'll do. Thanks for taking care of the cache error.

rfloca:
## SonarCloud findings

  The SonarCloud scan reported 7 new-code issues. I've fixed 2 and recommend marking the
  other 5 as **Won't fix**. Reason: they are either C++20-only suggestions (this project
  builds as **C++17**, `CMakeLists.txt:17`) or recommend changes that would break the
  existing public API / established code style -> out of scope for this PR.

  ### Fixed
  - **S7121** (redundant `c_str()`): `JSONSegmentationMetaInformationHandler` takes its
    argument as `std::string`, so passing `.c_str()` caused a needless
    `string → const char* → string` round-trip. Removed at both call sites.
  - **S6020** (`std::is_same<...>::value` → `std::is_same_v`): modernized all four
    occurrences so the string and handler overloads stay consistent.

  ### Won't fix (with rationale)
  - **S6190** — *"Use `std::source_location` instead of `__FILE__`/`__LINE__`."*
    `std::source_location` is C++20; this project is C++17. Independently of that, the
    flagged code is a test assertion **macro** that stringizes the asserted expression
    (`#expr`) and performs an early `return EXIT_FAILURE` from `main()` on failure —
    neither of which a free function can do. The macro form is required here.

  - **S6195** (×2) — *"Replace `enable_if` with a `requires` clause / abbreviated template
    / `if constexpr`."* `requires` clauses, concepts, and abbreviated function templates
    are C++20. `if constexpr` is a statement and cannot constrain template overload
    participation (SFINAE), so it cannot replace this constraint. On C++17,
    `std::enable_if_t` is the correct idiom, and it matches the pre-existing string
    overload's declaration. Worth revisiting if/when the project moves to C++20.

  - **S107** — *"Function has 8 parameters (> 7)."* The new handler overload deliberately
    mirrors the existing public `itkimage2dcmSegmentation(..., const string&, ...)`
    overload, which already carries this exact 8-parameter signature. The only way to
    satisfy this rule is to bundle the trailing flags into an options struct — a breaking
    API change affecting both overloads and every caller (MITK, the SEM CLI apps, tests).
    That is out of scope for this feature PR and would be better handled as a separate,
    dedicated refactor.

fedorov (approved):

PR #541, opened 2026-06-12
Use Pixel Padd. Value to designate labelmap bg.

### LINKED ISSUE #537: Do not force consecutive SegmentNumber assignment while writing LABELMAP
It does not appear such requirement is present: https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.4.html

michaelonken:
Following the discussion in Slicer/Slicer#9163, Pixel Padding Value (0028,0120) is the indicator that a labelmap segment is to be treated as background; a segment merely typed as Background (DCM,125040) is a regular segment. Requires DCMTK 5708ba6 (bumped), where the designated background is the single source of truth from which both the Background segment and Pixel Padding Value are derived on write.

Writing (ITK->labelmap, binary->labelmap):
- ConverterBase::addBackgroundSegmentIfNeeded() now designates pixel value 0 via DcmSegmentation::setBackgroundPixelValue() instead of hand-inserting a segment, so Pixel Padding Value 0 is written. Previously the eager insertion covered value 0 before DCMTK's write harmonization ran, so no Pixel Padding Value was ever written. The Background segment now carries the DCMTK default codes (category (SCT,309825002), type (DCM,125040), algorithm MANUAL); dcmqi still sets CIELab black resp. the Palette Color LUT entry.

Reading (labelmap->ITK):
- The segment designated by Pixel Padding Value gets no entry in the JSON segmentAttributes (pixel data is exported unchanged); warn if the designated value is non-zero. This removes the labelID 0 entry that shifted metadata assignment in Slicer.
- Slice positions not encoded as frames are filled with the Pixel Padding Value if present, 0 otherwise; if no Pixel Padding Value is present, frames are missing and a segment occupies pixel value 0, the conversion is refused (missing data would be indistinguishable from that segment).

Segment numbers (fixes #537):
- Labelmap output no longer enforces consecutive label IDs with --useLabelIDAsSegmentNumber; LABELMAP only requires uniqueness, so gaps are allowed. Label ID collisions across input files are rejected (labelmap addSegment() is an upsert and would silently drop a segment). Bit depth is chosen from the highest segment number instead of the segment count, and the binary->labelmap 8/16-bit threshold off-by-one (256 segments need 16 bits) is fixed.

Tests: JSON metadata roundtrip for labelmaps (fails if Pixel Padding Value is not written or not honored), gapped-label-ID write/pixel/JSON roundtrip, sparse-labelmap fill and refusal cases. New test data is described at the test definitions: relabeled copies of the spine and liver+spine NRRDs (label 2 -> 5), and two dcmodify-derived sparse 24x38x3 labelmaps with Pixel Padding Value 5 resp. none.


fedorov (commented):
Review by Claude.

---

## Overview

This PR switches labelmap background handling from a hand-inserted Background-typed segment to the DICOM-standard **Pixel Padding Value (0028,0120)** as the single source of truth, across three flows:

- **Write (ITK/binary → labelmap):** `addBackgroundSegmentIfNeeded()` now calls `DcmSegmentation::setBackgroundPixelValue(0)` instead of building/inserting a segment by hand, so Pixel Padding Value 0 actually gets written.
- **Read (labelmap → ITK):** the PPV-designated segment is omitted from `segmentAttributes`; missing (sparse) slices are filled with PPV (or 0), and the conversion is refused when the fill would be ambiguous.
- **Segment numbers (fixes #537):** gaps in label IDs are allowed for labelmap output; bit-depth is chosen from the highest segment number, and an 8/16-bit off-by-one is fixed.

Well-scoped and carefully documented — the inline comments, PR description, and test matrix (gapped IDs, sparse fill, refusal case, JSON roundtrip) all directly exercise the new behavior. Comments below are mostly minor.

## Correctness — looks right

- **Bit-depth off-by-one** (`Bin2Label.cpp`): `numSegments > 256` → `> 255` is correct — binary segments are 1..N, so N=256 overflows 8-bit (0..255). The matching `maxSegmentNumber > 255` in `Itk2DicomConverter.cpp` is consistent.
- **`maxSegmentNumber > 65535` guard** correctly rejects label IDs that can't fit a Uint16 segment number.
- **Collision rejection** for `useLabelIDAsSegmentNumber` + labelmap is a genuine bug-fix: `addSegment()` is an upsert, so without the check a duplicate label ID would silently drop a segment.
- **Sparse-fill logic** is sound. `hasMissingFrames = getNumberOfFrames() < m_imageSize[2]` only trips on *interior* gaps (border gaps shrink the computed extent instead), and the refusal path (no PPV + a segment occupies value 0) correctly avoids producing data indistinguishable from a real segment.
- **`FillBuffer` shortcut** (`if (pixel != fillValue)`) is correct given the buffer is pre-filled; for `fillValue==0` it relies on `allocateITKImage` zero-initializing, the same assumption the pre-existing code already made.

## Points worth addressing

- **Abbreviated DCMTK pin** (`CMakeExternals/DCMTK.cmake`): the hash changed from a full 40-char SHA to the 7-char `"5708ba6"`. Short hashes are less reproducible and can become ambiguous as the repo grows — recommend pinning the full SHA, matching the previous convention.
- **Copyright-header removal** (`Bin2Label.h` / `Bin2Label.cpp`): the OFFIS / Open Connections copyright + module header block is deleted. Presumably deliberate given the author, but removing a license/attribution header is worth calling out explicitly (or restoring) so it doesn't look accidental.
- **Warning wording** (`Dicom2ItkConverterLabel.cpp`): `getBackgroundSegmentNumber()` returns a segment number, but the non-zero warning prints it as a "pixel value." For labelmaps these coincide, so not wrong, just potentially confusing.

## Tests

Good coverage, and the binary fixtures are documented with full provenance (relabeled NRRDs, dcmodify-derived sparse DICOMs). Two non-blocking notes:
- `labelmap_meta_roundtrip` compares back against the *original input* `seg-example.json`, correctly relying on `comparejson.py` being key-order/whitespace tolerant — the test is only as strict as that comparator.
- The `WILL_FAIL` refusal test invokes the app target while the success tests use the `Test` target; intentional (it just needs a non-zero exit), but slightly inconsistent.

## Risk

- **Hard dependency on an unreleased DCMTK commit.** `setBackgroundPixelValue`, `getBackgroundSegmentNumber`, `isBackgroundSegment`, and the write-time `harmonizeLabelmapBackground` coverage scan all require the bumped DCMTK. Expected for this project (DCMTK is pinned), but flagging the coupling.
- **Behavior change for existing files:** the labelID-0 entry is no longer emitted in `segmentAttributes` (this is what fixes the Slicer metadata shift). Worth confirming no downstream consumer depended on the old labelID-0 row.

## Verdict

Clean, well-reasoned, well-tested change that aligns labelmap background handling with the DICOM standard and fixes #537. Suggest (1) restoring the full DCMTK SHA in the pin and (2) confirming the `Bin2Label` copyright-header removal is intentional before merge.

---
*This review was generated by Claude (Claude Code) and posted on behalf of @andrey.fedorov.*

fedorov:
> if no Pixel Padding Value is present, frames are missing and a segment occupies pixel value 0, the conversion is refused (missing data would be indistinguishable from that segment).

@michaelonken I think this is too radical. If the producer messes up the code or does not initialize padding value, the user won't be able to work around this failure. How about we either fill missing data with 0 anyway and log a warning, or add a flag to force this?

Issue #554, opened 2026-08-04
Common Instance Reference may omit additional source instances when one SEG frame references multiple instances


infinitus01:
## Summary

When `itkimage2segimage` maps more than one source DICOM instance to the same SEG frame through the per-frame geometry-matching path, all matching source instances appear in the frame-level `SourceImageSequence`, but only the first source instance appears in the top-level Common Instance Reference.

This behavior was reproduced using a patient-free synthetic fixture.

## Tested version

```text
dcmqi v1.5.6-2-gd9d2a14
```

## Synthetic setup

Two synthetic 2×2 CT Image Storage instances were created with identical:

- `StudyInstanceUID`
- `SeriesInstanceUID`
- `FrameOfReferenceUID`
- `ImagePositionPatient`
- `ImageOrientationPatient`
- pixel spacing
- image dimensions

The two instances have different `SOPInstanceUID` values.

Because their geometry is identical, both source instances match the same SEG frame when running with:

```text
referencesGeometryCheck=1
```

## Observed result

The behavior was reproduced independently 3/3 times.

### Frame-level references

The frame-level path contains both source instances A and B:

```text
PerFrameFunctionalGroupsSequence
  > DerivationImageSequence
    > SourceImageSequence
```

### Top-level Common Instance Reference

The top-level path contains only source instance A:

```text
ReferencedSeriesSequence
  > ReferencedInstanceSequence
```

Source instance B is absent from the top-level `ReferencedInstanceSequence`.

Additional observations:

- `itkimage2segimage` exits with code 0.
- A SEG object is written successfully.
- No warning about the incomplete Common Instance Reference set was observed.

## Expected result / question

My reading of DICOM PS3.3 C.12.2 is that the Common Instance Reference should include the SOP instances referenced elsewhere in the SEG object.

I would therefore expect both source instances A and B to be included in the top-level `ReferencedInstanceSequence`.

Could the maintainers confirm whether the current output is expected behavior or a defect?

## Suspected implementation path

The relevant path appears to permit multiple entries in `srcimgItems`, but the Common Instance Reference collection subsequently uses only:

```cpp
srcimgItems[0]
```

This may explain why the frame-level reference contains all matching instances while the top-level Common Instance Reference contains only the first one.

This is currently a source-level hypothesis, not a proposed authoritative fix.

## Possible correction approaches

I have not submitted a patch yet because I would first like to understand the maintainers' preferred correction approach:

[...]

PR #557, opened 2026-08-21
Factor source image handling out of the SEG converter and restructure it

### LINKED ISSUE #192: Mapping of the ITK image slices to DICOM instances should be factored out of SEG converter

michaelonken:
Refactoring of `itkimage2segimage` and `itkimage2paramap`, with no intended
change in behavior. It stands on its own, but is also the groundwork for
multiframe source support, which follows in a separate PR.

Fixes #192.

## Why

The slice-to-instance mapping and the code building the source image
references lived inline in `Itk2DicomConverter.cpp`, copy-pasted three times
within `itkimage2dcmSegmentation()` and a fourth time in `ParaMapConverter`.
That function had grown to ~840 lines, holding the whole conversion in one
body, and the duplication had started to drift apart between the copies.

## What changes

**A new `SourceImageIndex` component** (`include/dcmqi/SourceImageIndex.h`,
`libsrc/SourceImageIndex.cpp`) owns everything about the source images:
normalizing the datasets into a list of referenceable frames, mapping the
slices of the converted image onto them by geometry, and creating the
Derivation Image items plus the Common Instance Reference module from that
mapping. Both converters use it, which removes the duplication and lets the
PM converter drop its `CastImageFilter` workaround (the mapping now works on
`itk::ImageBase` geometry directly).

**Three components carved out of the SEG converter**, each with a concrete
reason rather than just moving lines:

- `PerFrameFunctionalGroups` owns the functional groups reused for every frame
  and enforces the protocol they need. The rule "clear the derivation image
  group after the frame was added, or the next frame inherits its references"
  was previously written out twice and left to caller discipline.
- `SegmentNumberAssigner` owns the segment numbering scheme and the mappings
  derived from it. The rules were restated in four places, including a second
  implementation of the labelmap bit depth inside the document creation.
- `SegmentationEncoder` turns the conversion into named steps
  (`createDocument()`, `addSharedFunctionalGroups()`, `mapSourceFrames()`,
  `encodeSegments()`, `encodeLabelmapFrames()`, `writeDocument()`). These share
  so much state that as free functions they would have needed a dozen
  parameters each; as members of one class they take none or few.

`itkimage2dcmSegmentation()` is now a 35-line wrapper. The public signatures
and the explicit template instantiations are unchanged.

## Behavior

Intended to be behavior-preserving, with four deliberate edge-case fixes in
the first commit, all in situations that previously produced questionable
output:

1. Frames whose slice has no mapped source no longer get an empty (and
   therefore invalid) `DerivationImageSequence` when other slices of the same
   input file do have references.
2. All instances referenced by a frame now reach `ReferencedSeriesSequence`;
   previously only the first one per frame did.
3. Labelmap frame bit depth now always matches the document bit depth. It
   could differ for gapped label IDs above 255.
4. Referenced instances are grouped by their actual series instead of all
   being attributed to the series of the first input dataset.

## Verification

- The existing test suite passes (76 tests).
- Each of the five commits was additionally checked by dcmdump-diffing 19
  conversion variants against a build of the preceding state: binary,
  labelmap, multiple input files, label IDs as segment numbers, gapped label
  IDs, overlapping segments, geometry check disabled, empty slices skipped and
  not skipped, and parametric maps. All byte-identical apart from UIDs and
  timestamps, including the console output.
- Because the later commits move error handling that none of those variants
  reach, the error paths were compared separately against a build of the
  original code: colliding label IDs across input files in both output
  flavors, a segment number beyond the 16 bit pixel range, and a label without
  metadata. Identical exit codes and identical messages.
- Compiles clean under `-Wall -Wextra -Wreorder`.

## Review order

The commits are meant to be read in order; each is self-contained and
individually verified. The first is the largest and carries the edge-case
fixes listed above.

### DIFF
--- a/apps/seg/Testing/CMakeLists.txt
+++ b/apps/seg/Testing/CMakeLists.txt
@@ -45,6 +45,33 @@ dcmqi_add_test(
     ${DICOM_DIR}/03.dcm
   )
 
+#-----------------------------------------------------------------------------
+# Regression test for the Common Instance Reference module
+# (https://github.com/QIICR/dcmqi/issues/554): when several source instances
+# map to the same segmentation frame, all of them must appear in the top-level
+# Referenced Instance Sequence, and instances must be grouped by the series
+# they actually belong to. The additional source instances (a same-plane
+# duplicate and one from a second series) are derived in memory from the
+# classic series, so no fixture files are needed.
+add_executable(CommonInstanceReferenceTest
+  CommonInstanceReferenceTest.cxx)
+target_link_libraries(CommonInstanceReferenceTest
+  dcmqi
+  ${DCMTK_LIBRARIES})
+set_target_properties(CommonInstanceReferenceTest PROPERTIES
+  LABELS ${MODULE_NAME})
+
+dcmqi_add_test(
+  NAME ${itk2dcm}_commonInstanceReference
+  MODULE_NAME ${MODULE_NAME}
+  COMMAND $<TARGET_FILE:CommonInstanceReferenceTest>
+    ${CMAKE_SOURCE_DIR}/doc/examples/seg-example.json
+    ${BASELINE}/liver_seg.nrrd
+    ${DICOM_DIR}/01.dcm
+    ${DICOM_DIR}/02.dcm
+    ${DICOM_DIR}/03.dcm
+  )
+
 dcmqi_add_test(
   NAME ${itk2dcm}_makeSEG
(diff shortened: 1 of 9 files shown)


michaelonken:
This PR also fixes #554 .

PR #559, opened 2026-09-02
Run dciodvfy conformance tests on macOS and Windows CI

fedorov:
## Problem

The `dciodvfy` conformance tests in `apps/{seg,sr,paramaps}/Testing/CMakeLists.txt` are
registered only if `find_program(DCIODVFY_EXECUTABLE dciodvfy)` succeeds. **No CI job has ever
provided `dciodvfy`, so all seven have been silently skipped everywhere**, on every platform,
since they were added.

The one place dicom3tools was nominally provisioned — `prereq.build_dicom3tools` in
`docker/Makefile` — does not work:

- the URL fetches `dicom3tools_winexe_*.zip`, i.e. **prebuilt Cygwin Windows binaries**, not source;
- it extracts that `.zip` with `tar xjf`, which GNU tar rejects (macOS bsdtar happens to accept
  it, which is why this went unnoticed);
- it then `cd`s into a versioned directory the flat archive does not contain.

The whole `&&` chain falls through to `|| echo "Reusing "`, so the step reports success while
producing nothing.

These tests are a genuine gate, not a formality: `dciodvfy` exits non-zero whenever it prints an
`Error`. Checked against 19 DICOM files under `data/`, exit status tracked error count exactly.

## Change

New composite action `.github/actions/install-dicom3tools` installs a prebuilt dicom3tools
distribution and prepends it to `PATH`. It is used from the macOS and Windows workflows, which
enables the seven tests on three jobs: `macos-15-intel`, `macos-15`, and `windows-2022`.

Design notes:

- **Digest pinning.** The archive's SHA-256 is verified, not just the release tag. A GitHub
  release asset can be replaced in place under an unchanged tag, so pinning the tag alone does
  not pin the bytes CI executes.
- **Loud failure.** The action smoke-tests that the binary can actually be *loaded* (exit >= 126
  means it could not be executed at all). Without this, an incompatible download would degrade
  straight back into the silent-skip behaviour this PR exists to fix.
- **Step placement.** The install runs early because `PATH` must be set before the *inner* dcmqi
  project is configured, which happens during the build step, not the configure step, of the
  superbuild.
- **Registration guard.** Each workflow asserts `ctest -N -R dciodvfy` finds a non-zero count
  before running the suite, so this cannot rot back into silence unnoticed.
- Extraction uses `7z` on Windows, as `unzip` is not present on the hosted Windows images.

## Linux is not covered, deliberately

The published Linux binaries are built on `ubuntu-22.04` and reference `GLIBC_2.34` /
`GLIBC_2.29`:

```
libc.so.6 -> GLIBC_2.2.5, GLIBC_2.4, GLIBC_2.14, GLIBC_2.34
libm.so.6 -> GLIBC_2.2.5, GLIBC_2.29
```

dcmqi's Linux tests run inside deliberately old-glibc containers, to keep the glibc floor of the
shipped binaries low:

| job | environment | glibc | usable? |
|---|---|---|---|
| `build-linux` | `dockcross/manylinux2014-x64` | 2.17 | no |
| `build-linux-arm64` | `quay.io/pypa/manylinux_2_28_aarch64` | 2.28 | no, and no aarch64 asset exists |

Both are below the floor, so the binary cannot be loaded. Requested upstream in
QIICR/dicom3tools equivalent repo: https://github.com/fedorov/dicom3tools/issues/19. Once
manylinux-built Linux assets exist, enabling Linux is one added step per job plus a digest bump.

Both Linux jobs are commented with this reasoning, replacing a comment that understated the
situation, and the broken `docker/Makefile` target is documented in place.

## Review notes

- **These jobs may go red on first run, and that is the point.** I could not verify locally that
  all seven tests pass on current `master` — reproducing needs a DCMTK rebuild, since dcmqi now
  uses `setBackgroundPixelValue` / `getBackgroundSegmentNumber`. Given `dciodvfy` has never
  gated anything here, and known conformance gaps already exist in dcmqi output (see the
  labelmap exclusion note in `apps/seg/Testing/CMakeLists.txt`), a latent error in a
  non-labelmap path is plausible. Worth watching this PR's own CI run.
- **Untested edge:** the Windows dicom3tools build is Cygwin-based, and it will receive
  Windows-style paths from ctest. Cygwin normally handles these, but it is not verified here.
- **`docker/Makefile` is left alone.** Its broken dicom3tools target is an independent bug;
  repairing or removing it is worth a separate change, since even a fixed version cannot use the
  current prebuilt Linux release.
- **Supply chain:** the action points at `fedorov/dicom3tools`, a personal repository. Digest
  pinning limits the blast radius, but if these binaries become a load-bearing CI dependency,
  moving the release to `QIICR/dicom3tools` — where the 2020 snapshot already lives — would be
  the more durable home.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/actions/install-dicom3tools/action.yml
+++ b/.github/actions/install-dicom3tools/action.yml
@@ -1,0 +1,89 @@
+name: "Install dicom3tools"
+description: >
+  Download a prebuilt dicom3tools distribution and prepend it to PATH, so that the
+  find_program(DCIODVFY_EXECUTABLE dciodvfy) calls in apps/*/Testing/CMakeLists.txt
+  succeed and the optional dciodvfy conformance tests are actually registered.
+
+# To move to a newer dicom3tools build, bump RELEASE_TAG and replace all three
+# digests below with the ones published for that release. The digests are verified
+# unconditionally: a release asset can be replaced in place under an unchanged tag,
+# so pinning the tag alone would not pin the bytes CI executes.
+
+runs:
+  using: composite
+  steps:
+    - name: "Install dicom3tools (dciodvfy)"
+      shell: bash
+      env:
+        RELEASE_REPO: fedorov/dicom3tools
+        RELEASE_TAG: dicom3tools.20260901072548.post1
+        SHA256_LINUX: d793a53a0f6cc4bae28b2979f67b583783d0954a7da7a2cd40b7359adbf681ae
+        SHA256_MACOS: a5d1f718a2d47f5dc5566ec34762bf55641b2f8c8c5c7d922404f5a6a9fe3afe
+        SHA256_WINDOWS: 9bb25b5ee3692cd2690c8f62007f2c352f5870c81c5c2ca3abf08042d44781da
+      run: |
+        set -euo pipefail
+
+        case "$RUNNER_OS" in
+          Linux)   asset=dicom3tools-linux-x86_64.tar.gz;     expected=$SHA256_LINUX ;;
+          macOS)   asset=dicom3tools-macos-universal.tar.gz;  expected=$SHA256_MACOS ;;
+          Windows) asset=dicom3tools-windows-x86_64.zip;      expected=$SHA256_WINDOWS ;;
+          *) echo "Unsupported runner OS: $RUNNER_OS" >&2; exit 1 ;;
+        esac
+
+        archive="$RUNNER_TEMP/$asset"
+        dest="$RUNNER_TEMP/dicom3tools"
+        mkdir -p "$dest"
+
+        url="https://github.com/$RELEASE_REPO/releases/download/$RELEASE_TAG/$asset"
+        echo "Downloading $url"
+        curl -fsSL --retry 3 --retry-delay 5 -o "$archive" "$url"
+
+        actual=$(openssl dgst -sha256 -r "$archive" | cut -d' ' -f1)
+        if [ "$actual" != "$expected" ]; then
+          echo "Checksum mismatch for $asset" >&2
+          echo "  expected: $expected" >&2
+          echo "  actual:   $actual" >&2
+          exit 1
+        fi
+
+        # All three archives are flat: every tool sits at the archive root, and the
+        # Windows build ships its Cygwin DLLs next to the executables, which is where
+        # Windows looks for them first.
+        case "$asset" in
+          *.tar.gz)
+            tar xzf "$archive" -C "$dest"
+            ;;
+          *.zip)
+            # unzip is not present on the hosted Windows images; 7z is.
+            if command -v 7z >/dev/null 2>&1; then
+              7z x -bso0 -o"$dest" "$archive"
+            else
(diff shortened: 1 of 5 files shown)


fedorov:
## First CI run: 5 of 7 tests pass, 2 caught a real bug

macOS arm64 result: `97% tests passed, 2 tests failed out of 77`. The three SEG and two SR
`dciodvfy` tests pass. Both failures are the parametric map ones, and they are a genuine
conformance defect, now fixed in 70589c3.

### "Warning" here really does fail validation

The output contains no `Error` line, only warnings, which makes the failure look spurious:

```
7: Warning - DimensionIndexPointer Attribute not present in any Shared or Per-Frame Functional Group - (0x0020,0x0032)
7: Warning - Missing attribute or value that would be needed to build DICOMDIR - Study ID
7: Warning - Value dubious for this VR - (0x0070,0x0084) PN Content Creator's Name  PN [1] = <dcmqi> - Retired Person Name form
7: ParametricMap
7: Warning - Unrecognized defined term <GEIIS> for value 1 of attribute <Coding Scheme Designator>
```

It is not the case that dciodvfy treats warnings as errors in general. `main` ends with
`return success ? 0 : 1`, and most warning sites deliberately leave `success` alone:

- the DICOMDIR warnings are called out in the source as *"always only warnings ... do not
  affect success"*;
- the unrecognized-defined-term warning is commented *"does NOT cause failure (cf.
  EnumValues)"*.

The first warning is the exception. Its emitting site in `appsrc/dcfile/dciodvfy.cc`
(`checkDimensionIndexValuesArePresentInFunctionalGroupsSequence`) prints a `W`-prefixed
message and then sets `success = false` immediately afterwards. So exactly one of these four
lines is fatal, and it is the one describing a real problem. The misleading severity wording
is a dicom3tools presentation issue, not a dcmqi one.

### The underlying defect

`ParaMapConverter.cpp` declared the dimension index as:

```cpp
mfdim.addDimensionIndex(DCM_ImagePositionPatient, dimUID,
                        DCM_RealWorldValueMappingSequence, "Frame position");
```

`FunctionalGroupPointer` must name the sequence that actually contains the attribute
`DimensionIndexPointer` refers to (PS3.3 C.7.6.17.1). `ImagePositionPatient` (0020,0032) is in
`PlanePositionSequence` (0020,9113), which `ParaMapConverter` already writes per frame via
`FGPlanePosPatient`. `RealWorldValueMappingSequence` (0040,9096) does not contain it, so
dciodvfy looked inside that sequence, found nothing, and correctly reported the dimension
index as dangling.

`Itk2DicomConverter.cpp` already passes `DCM_PlanePositionSequence` for the same attribute,
which is why only parametric map output was affected. Fixed by matching it.

No baseline or schema references the tag, so metadata round-trip comparisons are unaffected.

### Note for reviewers

This is the payoff from these tests never having run: the defect has presumably been shipping
in every parametric map dcmqi has produced. Worth deciding whether it deserves a CHANGELOG
entry under Fixed when the release notes are next generated.

